PR #54, opened 2022-01-14
fix(crab-strategy): price hedge time

alpinechicken:
# Task:

## Description

Added check that auction trigger time for price hedge is not before previous hedge timestamp. It was possible to specify any previous timestamp for the auction.

Fixes # 
Fixes auction so only possible to trigger after previous timestamp.

## Type of change

- [ ] Document update

## How Has This Been Tested

Test revert with timestamp before previous hedge that would otherwise have worked.

## Checklist

- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [x ] My changes generate no new warnings


antoncoding (approved):
need to run the linter with yarn lint

packages/hardhat/contracts/strategy/CrabStrategy.sol
```diff
@@ -578,6 +577,7 @@ contract CrabStrategy is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Ownab
         bool _isStrategySellingWSqueeth,
         uint256 _limitPrice
     ) internal {
+        require( _auctionTriggerTime > timeAtLastHedge, "Trigger time must be after time at last hedge");
```
aleone, line 580: This require is partially checked anytime we are doing a timeHedge via _isTimeHedge(): 
`       
 uint256 auctionTriggerTime = timeAtLastHedge.add(hedgeTimeThreshold);
`

So I think it might be cheaper/better to only check this         
`
require( _auctionTriggerTime > timeAtLastHedge, "Trigger time must be after time at last hedge");
`
in the isPriceHedge() check.  Thoughts?

aleone (approved):
I think it works as is, but think I suggested a better/cheaper way of doing it @haythem96 @antoncoding , thoughts?

packages/hardhat/contracts/strategy/CrabStrategy.sol
antoncoding: yeah good idea

haythem96:
I suggest also adding the require check in `_isTimeHedge()` and in `_isPriceHedge()` as that's where allowing to hedge logic is checked (and if the bug exist in hedge on uniswap function, this change will fix it)

haythem96:
Can u also refactor this line, not sure why I wrote this way different than the other functions: https://github.com/opynfinance/squeeth-monorepo/blob/fix/priceHedge-time/packages/hardhat/contracts/strategy/CrabStrategy.sol#L297

Change L297 and L299 to this https://github.com/opynfinance/squeeth-monorepo/blob/fix/priceHedge-time/packages/hardhat/contracts/strategy/CrabStrategy.sol#L328 and L330

aleone:
> I suggest also adding the require check in `_isTimeHedge()` and in `_isPriceHedge()` as that's where allowing to hedge logic is checked (and if the bug exist in hedge on uniswap function, this change will fix it)

Think we only need in _isPriceHedge(), but yeah agreed - see https://github.com/opynfinance/squeeth-monorepo/pull/54#pullrequestreview-854963746

alpinechicken:
> Can u also refactor this line, not sure why I wrote this way different than the other functions: https://github.com/opynfinance/squeeth-monorepo/blob/fix/priceHedge-time/packages/hardhat/contracts/strategy/CrabStrategy.sol#L297
> 
> Change L297 and L299 to this https://github.com/opynfinance/squeeth-monorepo/blob/fix/priceHedge-time/packages/hardhat/contracts/strategy/CrabStrategy.sol#L328 and L330

done

packages/hardhat/contracts/strategy/CrabStrategy.sol
```diff
@@ -824,6 +823,7 @@ contract CrabStrategy is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Ownab
      * @return true if hedging is allowed
      */
     function _isPriceHedge(uint256 _auctionTriggerTime) internal view returns (bool) {
+        require( _auctionTriggerTime > timeAtLastHedge, "Trigger time must be after time at last hedge");
```
aleone, line 826: @haythem96 is it bad for a view function to revert?

We expose checkPriceHedge which is an external view function.

aleone (approved):
just lintage

packages/hardhat/contracts/strategy/CrabStrategy.sol
haythem96: Yeah IMO `external view` should always return a result, in this case either True or False

haythem96: I guess to follow the initial design, u can add `if` statement inside `_isPriceHedge` in the beginning:
```
if (_auctionTriggerTime > timeAtLastHedge) return false;
```

aleone (approved):
lgtm

haythem96 (approved):
LGTM

Issue #42, opened 2022-01-13
Unrealized PnL for partially closed positions is off 


alexisgauba:
**Describe the bug**
- For partially closed positions, unrealized PnL should only take into account the part of the position that wasn't closed yet. Right now unrealized PnL takes into account the users full, initial position, not just what's left over after the partial close. That means unrealized PnL is too high for users after partial closes

**To Reproduce**

1. Open a long position 
2. Partially close it 
3. Compare the unrealized PnL

**Expected behavior**
After a partial close, unrealized PnL should only be based on the remaining amount

Issue #41, opened 2022-01-13
Take ETH collateral into account for short PnL 


alexisgauba:
Short PnL is currently calculated as `(amt of ETH earned from selling oSQTH) - (amt of ETH to pay to buy oSQTH back) * ETH price`, not taking ETH as collateral into account 

We should update it to `((amt of ETH earned from selling oSQTH) - (amt of ETH to pay to buy oSQTH back) * ETH price) + (ethDeposited * (ethPriceAtDeposit - ethPriceNow)) `

PR #58, opened 2022-01-14
Unrealized PnL for partially closed positions is off

EstherSeyi:
# Task:

## Description

This PR 

- updates short unrealized pnl calculations to account for a partially closed position.
- adds new long unrealized pnl calculation that will account for a partially closed position
- updates long unrealized pnl in position card and page

Fixes # (issue)
#42 
## Type of change

- [ ] Document update

## How Has This Been Tested

This PR can be tested by:

- Checking the calculations on the code to ensure that it's correct

## Checklist


packages/frontend/src/hooks/usePositions.ts
```diff
@@ -514,6 +514,7 @@ export const usePnL = () => {
   const currentShortDeposits = useMemo(() => {
     if (positionType === PositionType.LONG) return []
     let totalShortSqth = new BigNumber(0)
+    //@ts-expect-error
```
alexisgauba, line 517: hmm whats this?

EstherSeyi: Oh I forgot to remove it. It's a typescript error suppressor. I'll remove it

alexisgauba (commented):
Hmm this seems weird, it's giving me the same ETH PnL but different dollar PnLs on the deploy preview for this link vs. on squeeth.com 

Deploy preview: 
Screen Shot 2022-01-15 at 10 12 11 AM

Squeeth.com:
Screen Shot 2022-01-15 at 10 12 16 AM

EstherSeyi:
> Hmm this seems weird, it's giving me the same ETH PnL but different dollar PnLs on the deploy preview for this link vs. on squeeth.com
> 
> Deploy preview: Screen Shot 2022-01-15 at 10 12 11 AM
> 
> Squeeth.com: Screen Shot 2022-01-15 at 10 12 16 AM
I only changed the dollar one

alexisgauba:
Screen Shot 2022-01-15 at 3 51 40 PM
Hmm just getting 0 now, is it showing up for you?

EstherSeyi:
Yes it is, I will look into and fix

alexisgauba:
Hmm positions not showing up at all now 🤔 showing up for you? 

Screen Shot 2022-01-16 at 1 03 16 AM

EstherSeyi:
> https://continuouscall-git-fe-unrealized-pnl-partial-abb8b8-opynfinance.vercel.app

Yes, it is showing, I haven't changed anything that affects the display of the positions.

alexisgauba:
It's showing me different values on the positions card vs. the positions tab. Seems like the positions tab might be off, that looks way too high 

Screen Shot 2022-01-16 at 2 08 37 PM
Screen Shot 2022-01-16 at 2 08 58 PM

yichiehliu:
Umm found other issues, but can do it in the position fix pr later, the osquth-usd value on position card and position tab are different
image
image

yichiehliu (changes requested):
It's still showing wrong realized and unrealized pnl for long side in current vercel link 
current vercel link: 
image
squeeth.com:
image
The original one on squeeth.com seems to be closer to the correct number?
cuz 3874 x 0.0227 = 87.9389

EstherSeyi:
> https://continuouscall-git-fe-unrealized-pnl-partial-abb8b8-opynfinance.vercel.app



> It's still showing wrong realized and unrealized pnl for long side in current vercel link current vercel link: image squeeth.com: image The original one on squeeth.com seems to be closer to the correct number? cuz 3874 x 0.0227 = 87.9389

hmmm @yichiehliu in your opinion what will be the formula to derive this correct value?

packages/frontend/src/lib/pnl.ts
```diff
@@ -0,0 +1,95 @@
...
+  squeethAmount,
+  transactions,
+}: GetCurrentLongParams) {
+  if (positionType === PositionType.SHORT) return []
+  let totalLongSqth = new BigNumber(0)
+
+  const result: LongDeposit[] = []
+  let sellQuote
+  for (let index = 0; index < transactions.length; index++) {
+    if (totalLongSqth.gte(squeethAmount)) break
```
yichiehliu, line 39: Do we just break the loop here, or we need to do smt else if the amount is greater than squth amount?

packages/frontend/src/lib/pnl.ts
```diff
@@ -0,0 +1,95 @@
...
+    if (totalLongSqth.isLessThan(squeethAmount) && transactions[index].transactionType === TransactionType.BUY) {
+      totalLongSqth = totalLongSqth.plus(transactions[index].squeethAmount)
+      sellQuote = await getSellQuote(transactions[index].squeethAmount)
+      result.push({ ...transactions[index], sellQuote: sellQuote.amountOut })
+    } else if (
+      totalLongSqth.isLessThan(squeethAmount) &&
+      transactions[index].transactionType === TransactionType.SELL
+    ) {
+      totalLongSqth = totalLongSqth.minus(transactions[index].squeethAmount)
+    }
```
yichiehliu, line 49: What if the totalLongsqth is < squeethAmount because user transfer osqueeth to somewhere else instead of doing BUY or SELL?

packages/frontend/src/lib/pnl.ts
```diff
@@ -0,0 +1,95 @@
...
+}: GetCurrentShortParams) {
+  if (positionType === PositionType.LONG) return []
+  let totalShortSqth = new BigNumber(0)
+
+  const result: ShortDeposit[] = []
+  let buyQuote
+  for (let index = 0; index < transactions.length; index++) {
+    if (totalShortSqth.gte(squeethAmount)) break
+    if (
+      totalShortSqth.isLessThan(squeethAmount) &&
```
yichiehliu, line 80: User may mint and open short in two separate transactions, if we just check for MINT_SHORT and BURN_SHORT, then we will miss calculate some transaction
So does burn and buy back

yichiehliu (changes requested):

packages/frontend/src/lib/pnl.ts
yichiehliu: Like adding liquidity & removing liquidity and transfer osqueeth directly to other addresses

packages/frontend/src/lib/pnl.ts
yichiehliu: We can use the short helper does minting + selling in 1 tx vs 2 different tx for the mint on fe and sell on uniswap. There should be different addresses involved as well (Controller + uniswap pool directly from user address vs ShortHelper calling controller + uniswap from itself) -- quoted from Andrew lol

yichiehliu: MINT_SHORT is one tx, so does BURN_SHORT

yichiehliu: @alexisgauba wanna make sure that I understand positions correctly

yichiehliu: Also have to take adding & removing liquidity into account

packages/frontend/src/hooks/usePositions.ts
```diff
@@ -511,45 +513,55 @@ export const usePnL = () => {
...
-        totalShortSqth.isLessThan(squeethAmount) &&
-        transactions[index].transactionType === TransactionType.BURN_SHORT
-      ) {
-        totalShortSqth = totalShortSqth.minus(transactions[index].squeethAmount)
-      }
-    }
-    return result
-  }, [positionType, squeethAmount.toString(), transactions.length])
+  const currentShortDeposits = useAsyncMemo(
+    async () => await getCurrentShortDeposits({ positionType, squeethAmount, transactions, getBuyQuote }),
```
yichiehliu, line 517: Thinking we shld use finalSqueethAmount here instead of squeethAmount here, also need to take LPed amount into account, but there will be a loading issue, LPed amount is calculated in the last part, it will take a while to load

packages/frontend/src/lib/pnl.ts
alexisgauba: Yes people can go short and close short through the short helper but many people have also been doing it manually by minting and then selling on uniswap or buying back on uniswap and then burning

packages/frontend/src/hooks/usePositions.ts
alexisgauba: As long as we show a loading state, I think its more important to have the accurate value

packages/frontend/src/lib/pnl.ts
alexisgauba: Don't think adding / removing liquidity is needed here? We're just trying to show unrealized PnL for partially closed positions, so if someone closes their long partially or short partially

yichiehliu: But if we dont take them into account, will the squeeth amount be accurate?

packages/frontend/src/hooks/usePositions.ts
yichiehliu: The finalSqueeth should be the accurate value?

yichiehliu (changes requested):
The ratio seems to be off
image
image

packages/frontend/src/lib/pnl.ts
```diff
@@ -0,0 +1,43 @@
+import BigNumber from 'bignumber.js'
```
alexisgauba, line 1: would it makes sense to add this to utils instead of having a separate lib folder just for consistency or do you see lib as distinct from utils?

alexisgauba (commented):
Overall LGTM!

packages/frontend/src/lib/pnl.ts
EstherSeyi: Yes, I think utils is for more general things and a lib folder is good to have to keep calculations that are directly associated with each hook.

packages/frontend/src/hooks/usePositions.ts
```diff
@@ -120,7 +120,7 @@ export const usePositions = () => {
         totalUSDSpent: bigZero,
         ethCollateralPnl: bigZero,
       },
-    [isWethToken0, swaps?.length],
+    [ethPrice.toString(), isWethToken0, swaps?.length],
```
yichiehliu, line 123: Do we need to add ethPriceMap to dep array as well?

EstherSeyi: yh

packages/frontend/src/hooks/usePositions.ts
```diff
@@ -72,23 +72,32 @@ export const usePositions = () => {
           const wethAmt = new BigNumber(isWethToken0 ? s.amount0 : s.amount1)
           const usdAmt = getUsdAmt(wethAmt, s.timestamp)
 
+          const time = new Date(Number(s.timestamp) * 1000).setUTCHours(0, 0, 0) / 1000
+
+          acc.ethCollateralPnl = acc.ethCollateralPnl.plus(
+            wethAmt.abs().times(new BigNumber(ethPriceMap[time]).minus(ethPrice)),
```
yichiehliu, line 78: Do we accumulate this under all swaps? if we partially close the short position, it seems to accumulate the wethamount even I am trying to close position
I think we only need to accumulate eth collateral amount for selling txs?
Or maybe we use abs() instead of using negated()?

EstherSeyi: I actually think this works, it doesn't accumulate wethAmount when we partially close, it only calculates based on what's left in the position. Also, this calculation is only relevant when the user has a short position.

yichiehliu: But it seems to accumulate even when I have a bought swap, doesn't that mean to accumulate the eth collateral again when we partially close the position?

yichiehliu: Or even fully close the short position, for example, there will be two swaps, one is the selling tx, and the other is buying tx. We are going through both of them in this swaps iteration, but the eth collateral will accumulate only when we actually sell oSQTH. If we use abs() here, it means we make ethCollateral 2x bigger than it should be, cuz we accumulate it again

EstherSeyi: I'm not sure I follow. Are you saying we are not supposed to accumulate the ethcollateral for partially closed positions? If yes, why? @yichiehliu

yichiehliu: Like this two txs
`{
        "__typename": "Swap",
        "amount0": "16.240200606076410471",
        "amount1": "-5.339246529671475668",
        "id": "0xc1b340861c4c831fa5c1b271a2c54bfda33aaa0291359da87cd66f097e32ec7f#667",
        "origin": "0x5261ae032506ab52f6b2592b3135cb1b7741d205",
        "recipient": "0x8903918dfe74476e90b63061e5b9c3e63b65d3f4",
        "timestamp": "1642598425",
        "transaction": {
            "__typename": "Transaction",
            "blockNumber": "11842156",
            "id": "0xc1b340861c4c831fa5c1b271a2c54bfda33aaa0291359da87cd66f097e32ec7f"
        }
    },
    {
        "__typename": "Swap",
        "amount0": "-16.240200606076410471",
        "amount1": "5.361107043676061641",
        "id": "0xe6078c740536792b862ea5c7a9e78f91bbaa0226cd9fadcbc11a10ef15341f9a#675",
        "origin": "0x5261ae032506ab52f6b2592b3135cb1b7741d205",
        "recipient": "0x8903918dfe74476e90b63061e5b9c3e63b65d3f4",
        "timestamp": "1642613814",
        "transaction": {
            "__typename": "Transaction",
            "blockNumber": "11843124",
            "id": "0xe6078c740536792b862ea5c7a9e78f91bbaa0226cd9fadcbc11a10ef15341f9a"
        }
    },`

These two is where I open a short position and fully close it
We are going through both of them in this swaps iteration, but we will get wethAmt.abs()  = 5.361107043676061641 in the first tx and get 5.339246529671475668 in the second tx, plus it again, cuz we use abs() here

yichiehliu: and acc.ethCollateralPnl will be 5.361107043676061641 x price changes + 5.339246529671475668 x price changes, but the ethCollateral is only 5.361107043676061641

EstherSeyi: okay, I get it now @yichiehliu

yichiehliu: Also If they are just buying oSQTH, and then sell it
There wont be ethCollateralPnl at all since there is no ethCollateral in long scenario

yichiehliu:
Also If they are just buying oSQTH, and then sell it, there wont be ethCollateralPnl at all since there is no ethCollateral under long scenario

packages/frontend/src/hooks/usePositions.ts
yichiehliu: Also If they are just buying oSQTH, and then sell it, there wont be ethCollateralPnl at all since there is no ethCollateral under long scenario

packages/frontend/src/hooks/usePositions.ts
```diff
@@ -72,23 +72,33 @@ export const usePositions = () => {
...
-          acc.wethAmount = acc.wethAmount.plus(wethAmt.negated())
 
           //<0 means, buying squeeth
           //>0 means selling squeeth
           acc.totalSqueeth = acc.totalSqueeth.plus(squeethAmt)
           acc.totalETHSpent = acc.totalETHSpent.plus(wethAmt)
           acc.totalUSDSpent = acc.totalUSDSpent.plus(usdAmt)
 
-          acc.usdAmount = acc.usdAmount.plus(usdAmt)
+          if (squeethAmt.isPositive()) {
```
yichiehliu, line 86: it seems that the wethAmount from swap data is not equal to the ethCollateral amount
no matter when I open a new short position, or close it partially or close it fully, they are not equal to the eth collateral amount 
cc @alexisgauba

yichiehliu (changes requested):

packages/frontend/src/hooks/usePositions.ts
alexisgauba: Yep see #85

packages/frontend/src/lib/pnl.ts
```diff
@@ -1,21 +1,76 @@
...
+
+export function calcETHCollateralPnl(
+  data: VaultHistory_vaultHistories[] | undefined,
+  ethPriceMap: { [key: string]: number },
+  ethPrice: BigNumber,
+) {
+  return data?.length
+    ? data?.reduce((acc: BigNumber, curr: VaultHistory_vaultHistories) => {
+        const time = new Date(Number(curr.timestamp) * 1000).setUTCHours(0, 0, 0) / 1000
+        if (curr.action === 'DEPOSIT_COLLAT') {
```
yichiehliu, line 66: Just wondering if user uses MINT_SHORT to open a short position, do they also trigger the 'DEPOSIT_COLLAT' in vault history?

yichiehliu: Also if user uses WITHDRAW_COLLAT, the withdrawal amount of the collat shld be calculated in realized pnl instead of unrealized?

Also have a same q for BURN_SHORT, if user uses MINT_SHORT to open a short position, do they also trigger the 'WITHDRAW_COLLAT' in vault history?

yichiehliu: Besides, If the user gets liquidated, shld we also take the liquidated amount to get the new debt & collat and get the new realized & unrealized pnl? Does It depend on if it's a full liquidation or partial liquidation?

yichiehliu, line 66: Could use enum action type for DEPOSIT_COLLAT to avoid typo
```suggestion
        if (curr.action === Action.DEPOSIT_COLLAT) {
```

yichiehliu (changes requested):

packages/frontend/src/lib/pnl.ts
EstherSeyi: > Just wondering if user uses MINT_SHORT to open a short position, do they also trigger the 'DEPOSIT_COLLAT' in vault history?
> 
> Also have a same q for BURN_SHORT, if user uses MINT_SHORT to open a short position, do they also trigger the 'WITHDRAW_COLLAT' in vault history?

Yes it does you can actually test these by performing these operations and going on the subgraph to query vault history of your vault

EstherSeyi: > Besides, If the user gets liquidated, shld we also take the liquidated amount into account to get the new debt & collat and get the new realized & unrealized pnl? Does It depend on if it's a full liquidation or partial liquidation?

I don't know the answer to this.  Cc @alexisgauba @KMKoushik

packages/frontend/src/lib/pnl.ts
```diff
@@ -1,21 +1,76 @@
...
+          acc = acc.plus(
+            new BigNumber(toTokenAmount(curr.ethCollateralAmount, 18)).times(
+              new BigNumber(ethPriceMap[time]).minus(ethPrice),
```
yichiehliu, line 70: Shld be this?
```suggestion
             new BigNumber(ethPrice).minus(ethPriceMap[time]),
```

yichiehliu (changes requested):

packages/frontend/src/lib/pnl.ts
EstherSeyi: I don't think so, see [this](https://github.com/opynfinance/squeeth-monorepo/issues/41) issue.

yichiehliu: But for example if I have 1 ETH Collateral and the ethPriceAtDeposit is 100 and the ethPriceNow is 1000, the pnl would be 1 * (1000 - 100) = +900, or do I miss smt here

packages/frontend/src/lib/pnl.ts
alexisgauba: Full liquidation we don't need to take into account because that closes out their whole position. Partial liquidation would be good to take into account but imo doesn't need to be in this PR, can be later

packages/frontend/src/lib/pnl.ts
alexisgauba: yeah imo should be `ethPriceToClose - ethPriceAtDeposit`

packages/frontend/src/hooks/usePositions.ts
```diff
@@ -522,6 +537,19 @@ export const usePnL = () => {
...
   }
 
+  const { data } = useQuery<VaultHistory>(VAULT_HISTORY_QUERY, {
+    client: squeethClient[networkId],
+    fetchPolicy: 'cache-and-network',
+    variables: {
+      vaultId: shortVaults[firstValidVault]?.id,
+    },
+  })
+  const vaultHistoy = data?.vaultHistories
```
alexisgauba, line 547:
```suggestion
  const vaultHistory = data?.vaultHistories
```

packages/frontend/src/hooks/usePositions.ts
```diff
@@ -522,6 +537,19 @@ export const usePnL = () => {
...
+  const ethCollateralPnl = useMemo(
+    () => calcETHCollateralPnl(vaultHistoy, ethPriceMap, ethPrice),
```
alexisgauba, line 549:
```suggestion
    () => calcETHCollateralPnl(vaultHistory, ethPriceMap, ethPrice),
```

packages/frontend/src/hooks/usePositions.ts
```diff
@@ -522,6 +537,19 @@ export const usePnL = () => {
...
+    [vaultHistoy?.length, ethPrice, ethPriceMap],
```
alexisgauba, line 550:
```suggestion
    [vaultHistory?.length, ethPrice, ethPriceMap],
```

packages/frontend/src/lib/pnl.ts
```diff
@@ -1,21 +1,93 @@
...
-    !ethPrice.isFinite()
+    !ethPrice.isFinite() ||
+    ethCollateralPnl.isEqualTo(0) ||
+    !ethCollateralPnl.isFinite()
   ) {
     return new BigNumber(0)
   }
-  return wethAmount.minus(buyQuote).multipliedBy(ethPrice)
+
+  return wethAmount.minus(buyQuote).multipliedBy(ethPrice).plus(ethCollateralPnl)
```
alexisgauba, line 28:
```suggestion
  return buyQuote.minus(wethAmount).multipliedBy(ethPrice).plus(ethCollateralPnl)
```

packages/frontend/src/lib/pnl.ts
```diff
@@ -1,21 +1,93 @@
...
+      }, new BigNumber(0))
+    : new BigNumber(0)
+
+  const withdrawals = data?.length
+    ? data?.reduce((acc: BigNumber, curr: VaultHistory_vaultHistories) => {
+        const time = new Date(Number(curr.timestamp) * 1000).setUTCHours(0, 0, 0) / 1000
+        if (curr.action === Action.WITHDRAW_COLLAT) {
+          acc = acc.plus(
+            new BigNumber(toTokenAmount(curr.ethCollateralAmount, 18)).times(
+              new BigNumber(ethPriceMap[time]).minus(ethPrice),
```
alexisgauba, line 84:
```suggestion
              new BigNumber(ethPrice).minus(ethPriceMap[time]),
```

alexisgauba (changes requested):

packages/frontend/src/lib/pnl.ts
EstherSeyi: This was the part of the formula that @KMKoushik corrected the other day. That's why I changed it. See https://opyn.slack.com/archives/C02R5NAT6BA/p1642492674016200

alexisgauba (approved):

packages/frontend/src/lib/pnl.ts
KMKoushik: Liquidations are tough we can add it later

KMKoushik (approved):

Issue #62, opened 2022-01-16
Positions don't reload on the positions tab or vault manager 


alexisgauba:
**Describe the bug**
The positions card and short input are now showing the right value for positions, but it takes those about three reloads to show the right value. The positions tab and vault manager _don't_ reload, so positions doesn't show up at all and vault manager gets stuck showing the wrong value. This is potentially because the LP loading is not taken into account? 

<img width="1374" alt="Screen Shot 2022-01-15 at 6 08 04 PM" src="https://user-images.githubusercontent.com/13340486/149645615-cbb25278-bb84-48b4-ac61-711c09336517.png">
<img width="1389" alt="Screen Shot 2022-01-15 at 6 07 52 PM" src="https://user-images.githubusercontent.com/13340486/149645617-973b64ce-4dff-4d5e-ab2c-9a47952e51e1.png">
<img width="1189" alt="Screen Shot 2022-01-15 at 6 07 15 PM" src="https://user-images.githubusercontent.com/13340486/149645618-f97b6824-9c25-4f09-a88c-3e109faaac16.png">

**To Reproduce**
1. Mint some oSQTH 
2. Add liquidity to LP
3. Go short 
3. Remove LP liquidity 
4. Check positions card, short / close, positions tab, and vault manager

**Expected behavior**
- Positions tab should show a short position with the same value as positions card 
- Vault manager should show the same value as positions card for short debt, and should actually show the minted debt (right now shows 0)
- Adjust debt max to burn should show the oSqueethBal, not 0

yichiehliu:
Also fixing on collateral ratio indifference on vault manager and position page
![image](https://user-images.githubusercontent.com/26922518/150107758-757bca9c-f5a7-470b-927a-7b6af3910487.png)
![image](https://user-images.githubusercontent.com/26922518/150107872-587beb9c-76f9-4dfc-86c0-f6c638304eff.png)

PR #76, opened 2022-01-19
fix: position correct amount & vault manager debt amount & usePosition hook

yichiehliu:
# Task:

## Description

Fix:
1. Showed correct values for short squeeth, minted squeeth, LPed squeeth
2. Moved custom debt logic from vault manager, positions, and short component to usePositions
3. Changed to swap subscription in usePosition to update data fast enough for user, and dont need to refresh to see their positions
4. Fixed BN Object in dep array issue
5. Changed oSqthBal to longSqthBal, which is user osquth balance subtracted by minted oSQTH amount, for getting correct long balance in long trade ticket

Feat:
1. Added LPed amount on vault manager page
image
Fixes # (issue) #62 

## Type of change

- [ ] Document update

## How Has This Been Tested
Locally

## Checklist

- [x] Added video recordings if it is a UI change


packages/frontend/src/hooks/usePositions.ts
```diff
@@ -41,7 +42,28 @@ export const usePositions = () => {
...
+    subscribeToNewPositions()
+  }, [])
+
+  const subscribeToNewPositions = useCallback(() => {
+    subscribeToMore({
+      document: SWAPS_SUBSCRIPTION,
+      variables: {
+        poolAddress: squeethPool?.toLowerCase(),
+        origin: address || '',
+        recipients: [shortHelper, address || '', swapRouter],
```
alexisgauba, line 55: this adjust the query to get swaps from short helper and the Uniswap swap router for individual sells after mints?

packages/frontend/src/hooks/usePositions.ts
```diff
@@ -112,12 +134,33 @@ export const usePositions = () => {
...
+  const lpDebt = useMemo(() => {
+    return depositedSqueeth.minus(withdrawnSqueeth).isGreaterThan(0)
+      ? depositedSqueeth.minus(withdrawnSqueeth)
+      : new BigNumber(0)
+  }, [positionType, depositedSqueeth.toString(), withdrawnSqueeth.toString()])
+
   const { finalSqueeth, finalWeth } = useMemo(() => {
-    const finalSqueeth = squeethAmount.minus(depositedSqueeth).plus(withdrawnSqueeth)
+    // dont include LPed amount will be the correct short amount
+    const finalSqueeth = squeethAmount
```
alexisgauba, line 159: wdym correct short amount? If you are just looking for how much someone is short that should not include both minted and LPed

alexisgauba (commented):
Overall looking really good! 

Just a few things 
- Would seperate out LP 
- Looks like PnL is showing NaN now 
Screen Shot 2022-01-19 at 5 22 29 PM

packages/frontend/src/hooks/usePositions.ts
yichiehliu: Oh, this adjust is for subscription, I didn't change the swaps pulling logics

packages/frontend/src/hooks/usePositions.ts
yichiehliu: But the squeethAmount is already the short amount that someone has, the LPed and minted amount was not accounted as squeethAmount in the beginning so we dont need to subtract LPed and minted amount here

yichiehliu:
Right now it should be able to separate out LPed amount from minted amount, it doesn't work for you?
image

alexisgauba:
> Right now it should be able to separate out LPed amount from minted amount, does it work for you? image

On the vault manager it groups minted and LP together which isn't accurate, since they are both very distinct positions

packages/frontend/src/hooks/usePositions.ts
alexisgauba: Ok don't we need to adjust the swap query to make sure we get all Uniswap swaps not just ones from our short helper?

packages/frontend/src/hooks/usePositions.ts
alexisgauba: Ok, comment was confusing because it only called out LP not minting as well

yichiehliu: lol gotcha, will edit this

EstherSeyi:
Looks good to me asides from the things that have already been pointed out.

packages/frontend/src/hooks/usePositions.ts
yichiehliu: but do we really miss some txs here?

alexisgauba:
Screen Shot 2022-01-20 at 10 05 46 AM
we'd need to add back showing LP squeeth on the vault manager as well!

packages/frontend/src/hooks/usePositions.ts
alexisgauba: yeah if a user buys back on Uniswap directly it won't capture that right? Or if they just mint and then go sell

yichiehliu: Ah kk, maybe I  fix this issue in another PR?

alexisgauba (approved):
lgtm!

KMKoushik (approved):

Issue #46, opened 2022-01-13
Site is being slow, investigate potential memory leak


alexisgauba:
**Describe the bug**
- Site is giving user a slow loading error, indicating potential memory 

**Screenshots**
<img width="678" alt="Screen Shot 2022-01-12 at 11 00 47 PM" src="https://user-images.githubusercontent.com/13340486/149281386-0bedb820-610e-46c2-a069-6b0e9b0782d4.png">

alexisgauba:
Additionally Koushik's suggestion on speed: Just realised useETHprice and useTokenBalance is used in multiple places. So in a page there can be 5-6 requests every 10 seconds. We can move this to a context and use it from there. This will drastically reduce the calls. And instead of polling at time interval we can poll at new blocks

PR #86, opened 2022-01-21
Fe/memory leak

EstherSeyi:
# Task: Site is being slow, investigate a potential memory leak

## Description

This PR basically tries to refactor code across the app in a bid to improve the performance. 

Fixes # (issue)
#46 
## Type of change

- [ ] Document update

## How Has This Been Tested

This PR can be tested by initiating transactions and looking out for the site giving a slow loading error.
There seems to still be a memory leak error on the Trade ticket which I have been unable to pinpoint the cause (opened to suggestions) but overall, performance is improved.
## Checklist


yichiehliu (commented):
could u update to the latest changes?

EstherSeyi:
> Could u update to the latest changes?

okay, will do

packages/frontend/src/components/Trade/Long/index.tsx
```diff
@@ -532,9 +522,9 @@ const CloseLong: React.FC<BuyProps> = ({
       if (longSqthBal.lt(amount)) {
         closeError = 'Insufficient oSQTH balance'
       }
-      if (amount.gt(balance)) {
-        openError = 'Insufficient ETH balance'
-      }
+      // if (amount.gt(balance)) {
```
alexisgauba, line 525: why comment this out?

packages/frontend/src/context/wallet.tsx
```diff
@@ -101,26 +100,21 @@ const WalletProvider: React.FC = ({ children }) => {
       disconnectWallet: disconnectWallet,
       handleTransaction,
     }),
-    [web3, address, networkId, signer, balance, onWalletSelect, disconnectWallet],
+    [web3, address, networkId, signer, balance.toString()],
   )
```
alexisgauba, line 105: not able to connect my wallet from the deploy preview - is that happening for you? Everytime I click "connect wallet" I get this: 

Screen Shot 2022-01-22 at 2 01 56 PM

packages/frontend/src/components/Trade/Long/index.tsx
```diff
@@ -243,15 +241,15 @@ const OpenLong: React.FC<BuyProps> = ({ balance, open, setTradeCompleted, active
...
+  // let closeError: string | undefined
   let existingShortError: string | undefined
   let priceImpactWarning: string | undefined
 
   useEffect(() => {
     if (connected) {
-      if (longSqthBal.lt(amount)) {
-        closeError = 'Insufficient oSQTH balance'
-      }
+      // if (oSqueethBal.lt(amount)) {
```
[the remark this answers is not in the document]
EstherSeyi: yes i think so, will compare to production and correct

packages/frontend/src/components/Trade/Long/index.tsx
EstherSeyi: openError is not being used in this component

packages/frontend/src/context/wallet.tsx
EstherSeyi: Yes this is happening to me, will update

packages/frontend/pages/index.tsx
```diff
@@ -518,7 +518,7 @@ function TradePage() {
 
   useEffect(() => {
     setVolMultiplier(1.2)
-  }, [tradeType])
+  }, [])
```
alexisgauba, line 521: why remove this dep?

alexisgauba (commented):
other than that one dep, lgtm!

packages/frontend/pages/index.tsx
EstherSeyi: It seems like it's not being used since the volMultiplier is always being set to 1.2 here

alexisgauba: if there's nothing in the dependency array does that mean it will always re-render or am I mistaken? Believe it was originally there because the Short side may have had a different vol multiplier, but we don't show a short side backtest right now 

We actually might not need this setter at all, think in the long backtest code EJ has it set at 1.2, but cc @yichiehliu to confirm?

EstherSeyi: No, it will only be set with the initial render if there's nothing in the array. But I removed it now after asking @yichiehliu.

alexisgauba (approved):
Lgtm!

yichiehliu (approved):

PR #154, opened 2022-02-08
Fe/direct selling swaps

### LINKED ISSUE #87: Update swaps query to pull from the shortHelper and Uniswap directly
**Describe the bug**
Right now the swaps query on the short side doesn't consider the Uniswap router, when it should because users manually mint and then sell on Uniswap, or manually buy back on Uniswap to burn 

**To Reproduce**
Manually mint and sell on Uniswap 

**Expected behavior**
Swaps router takes into account txns on Uniswap and properly classifies them as short if mint + sell directly on Uniswap

yichiehliu:
# Task:

## Description

1. add direct selling txs to txs history and make the sold balance correct if users manually sell oSQTH on uniswap(since currently manual selling is only supported on mainnet, would use original query method when users is not on the mainnet)

Fixes #87  (issue)

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/packages/frontend/src/hooks/useTransactionHistory.ts
+++ b/packages/frontend/src/hooks/useTransactionHistory.ts
@@ -93,7 +93,7 @@ export const useTransactionHistory = () => {
     const usdValue = ethAmount.multipliedBy(ethPriceMap[time]).abs()
 
     let transactionType = TransactionType.BUY
-    if (squeethAmount.isPositive() && s.recipient.toLowerCase() === swapRouter.toLowerCase()) {
+    if (squeethAmount.isPositive() && s.recipient.toLowerCase() !== shortHelper.toLowerCase()) {
       transactionType = TransactionType.SELL
     } else if (s.recipient.toLowerCase() === shortHelper.toLowerCase()) {
       if (squeethAmount.isNegative()) transactionType = TransactionType.BURN_SHORT
--- a/packages/frontend/src/hooks/useTransactionHistory.ts
+++ b/packages/frontend/src/hooks/useTransactionHistory.ts
@@ -4,11 +4,12 @@ import BigNumber from 'bignumber.js'
 import { TransactionType } from '../constants'
 import { useWallet } from '@context/wallet'
 import { useWorldContext } from '@context/world'
-import { transactions_swaps, transactions_positionSnapshots } from '../queries/uniswap/__generated__/transactions'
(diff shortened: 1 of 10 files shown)


packages/frontend/src/hooks/useTransactionHistory.ts
```diff
@@ -84,14 +86,14 @@ export const useTransactionHistory = () => {
...
-  const transactions = (data?.swaps || []).map((s: transactions_swaps) => {
+  const transactions = (swaps || []).map((s) => {
     const squeethAmount = new BigNumber(isWethToken0 ? s.amount1 : s.amount0)
     const ethAmount = new BigNumber(isWethToken0 ? s.amount0 : s.amount1)
     const time = new Date(Number(s.timestamp) * 1000).setUTCHours(0, 0, 0) / 1000
     const usdValue = ethAmount.multipliedBy(ethPriceMap[time]).abs()
 
     let transactionType = TransactionType.BUY
-    if (squeethAmount.isPositive() && s.recipient.toLowerCase() === swapRouter.toLowerCase()) {
+    if (squeethAmount.isPositive() && s.recipient.toLowerCase() !== shortHelper.toLowerCase()) {
```
alexisgauba, line 96: hm not sure I understand this, we also want to include txns that go through shortHelper right? Or are we not tracking them here because we are tracking them elsewhere and this is just meant to be for swapRouter txns?

yichiehliu: There are two conditions here, if s.recipient.toLowerCase() === shortHelper.toLowerCase(), that means users are doing OPEN_SHORT or CLOSE_SHORT, and if s.recipient.toLowerCase() !== shortHelper.toLowerCase(), that means users are doing BUY or SELL, but the pattern wont work for minted/burnt only txs, will figure out a way to distinguish them in the following pr.

alexisgauba: Ok gotcha so this only affects the txn history, not the positions stuff, and it's just showing the distinction between open_short vs. minting vs. selling? 

And the next PR you're adding the separate actions?

yichiehliu: Yh after this pr I will add separate actions, cuz it requires new vault history to distinguish minting from MINT_SHORT

alexisgauba (approved):

Issue #156, opened 2022-02-09
Positions page shows short position even after fully closed out 


alexisgauba:
**Describe the bug**
Opened and closed a short position, but the positions tab still thought I was short even when I had fully closed out the short position 

I did mint after I closed out, but the full correct minted amount is showing up in the minted section.

**To Reproduce**

1. Open a short position 
2. Close a short position

**Expected behavior**
Should not have shown any short position

**Screenshots**
- Mint and sell: https://etherscan.io/tx//0x542092a77a02f0286575b75522baf02b449c17643e32b53e40905315fab59549
- Buy back and burn: https://etherscan.io/tx//0x162e3e0fa0a21683072b089b1b2c0a4a67cf11b98951891f4638297b1c0930ad

<img width="1060" alt="Screen Shot 2022-02-08 at 6 25 14 PM" src="https://user-images.githubusercontent.com/13340486/153110916-43ff2756-48da-418b-9c5f-8ffbad3996c0.png">

PR #159, opened 2022-02-10
Fixes positions page showing short position even after fully closed out

EstherSeyi:
# Task:

## Description

This PR fixes positions page showing short position even after fully closed out

Fixes # (issue)
#156 
## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/packages/frontend/src/components/PositionCard.tsx
+++ b/packages/frontend/src/components/PositionCard.tsx
@@ -195,6 +195,7 @@ const PositionCard: React.FC<PositionCardType> = ({ tradeCompleted }) => {
 
   useEffect(() => {
     if (tradeSuccess && prevSwapsData?.length === swaps?.length) {
+      //if trade success and number of swaps is still the same, try refetching again
       swapsQueryRefetch()
     } else {
       setTradeSuccess(false)


packages/frontend/src/components/PositionCard.tsx
```diff
@@ -14,6 +14,7 @@ import { useTrade } from '@context/trade'
 import { useWorldContext } from '@context/world'
 import { PositionType, TradeType } from '../types'
 import { useVaultLiquidations } from '@hooks/contracts/useLiquidations'
+import { usePrevious } from 'react-use'
```
alexisgauba, line 17: what does this hook do / how is it working under the hood?

EstherSeyi: It stores the previous value of a state. I realized that the issue was that the swaps data was set up to refetch 5secs after a successful trade but sometimes the refetch was not successful so it won't update the UI. So now I compare the previous swap state after a trade success to know if the state has been updated or not. 

Under the hood, it uses useRef hook for storing the previous state value. see [here](https://github.com/streamich/react-use/blob/master/src/usePrevious.ts)

alexisgauba (approved):

PR #173, opened 2022-02-16
feat: flash mint

haythem96:
# feat: flash mint

## Description

This PR add the flash mint functionality, where the helper contract flashswap the amount of ETH collateral needed to mint from uniswap pool, and repay it with the minted wPowerPerp amount.

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/packages/hardhat/contracts/periphery/ControllerHelper.sol
+++ b/packages/hardhat/contracts/periphery/ControllerHelper.sol
@@ -39,12 +39,12 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
 
     struct flashswapWMintData {
         uint256 vaultId;
-        uint256 flashSwapedCollateral;
+        uint256 flashSwappedCollateral;
         uint256 totalCollateralToDeposit;
         uint256 wPowerPerpAmount;
     }
 
-    event flashswapWMint(
+    event FlashswapWMint(
         address indexed depositor,
         uint256 vaultId,
         uint256 wPowerPerpAmount,
@@ -106,7 +106,7 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
             abi.encodePacked(_vaultId, amountToFlashswap, _collateralAmount, _wPowerPerpAmount)
         );
 
-        emit flashswapWMint(msg.sender, _vaultId, _wPowerPerpAmount, amountToFlashswap, _collateralAmount);
+        emit FlashswapWMint(msg.sender, _vaultId, _wPowerPerpAmount, amountToFlashswap, _collateralAmount);
     }
 
     /**
@@ -132,7 +132,7 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
             // convert WETH to ETH as Uniswap uses WETH
             IWETH9(weth).withdraw(IWETH9(weth).balanceOf(address(this)));
 
-            //will revert if data.flashSwapedCollateral is > eth balance in contract
+            //will revert if data.flashSwappedCollateral is > eth balance in contract
             // IController(controller).mintWPowerPerpAmount{value: address(this).balance}(data.vaultId, data.wPowerPerpAmount, 0);
             uint256 vaultId = IController(controller).mintWPowerPerpAmount{value: data.totalCollateralToDeposit}(
                 data.vaultId,
--- a/packages/hardhat/contracts/periphery/ControllerHelper.sol
+++ b/packages/hardhat/contracts/periphery/ControllerHelper.sol
@@ -37,14 +37,14 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
     address public immutable wPowerPerp;
     address public immutable weth;
 
-    struct FlashWMintData {
+    struct flashswapWMintData {
         uint256 vaultId;
         uint256 flashSwapedCollateral;
         uint256 totalCollateralToDeposit;
         uint256 wPowerPerpAmount;
     }
 
-    event FlashWMint(
+    event flashswapWMint(
         address indexed depositor,
         uint256 vaultId,
         uint256 wPowerPerpAmount,
@@ -89,7 +89,7 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
         require(msg.sender == weth || msg.sender == address(controller), "Cannot receive eth");
     }
 
-    function flashWMint(
+    function flashswapWMint(
         uint256 _vaultId,
         uint256 _wPowerPerpAmount,
         uint256 _collateralAmount
@@ -106,7 +106,7 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
             abi.encodePacked(_vaultId, amountToFlashswap, _collateralAmount, _wPowerPerpAmount)
         );
 
-        emit FlashWMint(msg.sender, _vaultId, _wPowerPerpAmount, amountToFlashswap, _collateralAmount);
+        emit flashswapWMint(msg.sender, _vaultId, _wPowerPerpAmount, amountToFlashswap, _collateralAmount);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


packages/hardhat/test/integration-tests/periphery/controller-helper.ts
```diff
@@ -0,0 +1,135 @@
...
+      await controllerHelper.connect(depositor).flashWMint(0, mintWSqueethAmount, collateralAmount, {value: value});
+
+      const controllerBalanceAfter = await provider.getBalance(controller.address)
+      const squeethBalanceAfter = await wSqueeth.balanceOf(depositor.address)
+      const vaultAfter = await controller.vaults(vaultId)
+
+      // Check why this failing with small diff
+      // expect(controllerBalanceBefore.add(collateralAmount).eq(controllerBalanceAfter)).to.be.true
+      expect(squeethBalanceBefore.eq(squeethBalanceAfter)).to.be.true
+      // expect(vaultBefore.collateralAmount.add(collateralAmount).eq(vaultAfter.collateralAmount)).to.be.true
```
aleone, line 131: this is failing because of the fee

aleone: can either remove fee or calculate expected fee and subtract it from collateral in controller and collateral in vault

packages/hardhat/test/integration-tests/periphery/controller-helper.ts
```diff
@@ -0,0 +1,135 @@
...
+    await provider.send("evm_mine", [])
+  })
+
+  describe("Mint short with flash deposit", async () => {
+    it("flash mint", async () => {      
+      const vaultId = await shortSqueeth.nextId();
+      // await controller.connect(depositor).updateOperator(vaultId, controllerHelper.address)
+
+      const normFactor = await controller.normalizationFactor()
+      const mintWSqueethAmount = ethers.utils.parseUnits('2')
```
aleone, line 111: think should make this larger as we will have to update tests to take into account 6.9 eth min collateral (might as well make it work now)

packages/hardhat/test/integration-tests/periphery/controller-helper.ts
```diff
@@ -0,0 +1,135 @@
...
+      const ethPrice = await oracle.getTwap(ethDaiPool.address, weth.address, dai.address, 420, true)
+      const scaledEthPrice = ethPrice.div(10000)
+      const debtInEth = mintRSqueethAmount.mul(scaledEthPrice).div(one)
+      const collateralAmount = debtInEth.mul(3).div(2).add(ethers.utils.parseUnits('0.01'))
+      const value = collateralAmount.div(2).add(collateralAmount.div(3))
+      const controllerBalanceBefore = await provider.getBalance(controller.address)
+      const squeethBalanceBefore = await wSqueeth.balanceOf(depositor.address)
+      const vaultBefore = await controller.vaults(vaultId)
+
+      await controllerHelper.connect(depositor).flashWMint(0, mintWSqueethAmount, collateralAmount, {value: value});
```
aleone, line 122: think we should test this in a way where we set a slippage and ensure it matches

packages/hardhat/test/integration-tests/periphery/controller-helper.ts
```diff
@@ -0,0 +1,135 @@
...
+      // await controller.connect(depositor).updateOperator(vaultId, controllerHelper.address)
+
+      const normFactor = await controller.normalizationFactor()
+      const mintWSqueethAmount = ethers.utils.parseUnits('2')
+      const mintRSqueethAmount = mintWSqueethAmount.mul(normFactor).div(one)
+      const ethPrice = await oracle.getTwap(ethDaiPool.address, weth.address, dai.address, 420, true)
+      const scaledEthPrice = ethPrice.div(10000)
+      const debtInEth = mintRSqueethAmount.mul(scaledEthPrice).div(one)
+      const collateralAmount = debtInEth.mul(3).div(2).add(ethers.utils.parseUnits('0.01'))
+      const value = collateralAmount.div(2).add(collateralAmount.div(3))
```
aleone, line 117: I think to calculate value we should:
- pick a oSQTH amount
- figure out the CR we want (say 150% or 200% or whatever)
- determine the amount of collateral that corresponds to 
- determine the uniswap price
- figure out how much ETH the set oSQTH will be sold for + some set slippage
- figure out what msg.value needs to be from collateral - expectedEthProceeds * (1-slippage)

aleone (changes requested):
- few suggested changes on way to test
- should test slippage and ensure its correct
- ensure refund is roughly slippage
- fix fee calculation (remove or take into account)

packages/hardhat/test/integration-tests/periphery/controller-helper.ts
```diff
@@ -105,30 +105,31 @@ describe("Controller helper integration test", function () {
...
-      const mintWSqueethAmount = ethers.utils.parseUnits('2')
+      const mintWSqueethAmount = ethers.utils.parseUnits('10')
       const mintRSqueethAmount = mintWSqueethAmount.mul(normFactor).div(one)
       const ethPrice = await oracle.getTwap(ethDaiPool.address, weth.address, dai.address, 420, true)
       const scaledEthPrice = ethPrice.div(10000)
       const debtInEth = mintRSqueethAmount.mul(scaledEthPrice).div(one)
       const collateralAmount = debtInEth.mul(3).div(2).add(ethers.utils.parseUnits('0.01'))
-      const value = collateralAmount.div(2).add(collateralAmount.div(3))
+      const squeethEthPrice = await oracle.getTwap(wSqueethPool.address, wSqueeth.address, weth.address, 420, true)
+      const ethAmountOut = mintWSqueethAmount.mul(squeethEthPrice).div(one)
```
aleone, line 117: i still think its worth checking the exact expected out here (ie including price impact on uniswap), but lmk what you think

that way we can compare and ensure that the user is refunded value - ethAmountOut, which means that 0 of the slippage was used (which is to be expected when testing)

aleone (approved):
I still think should calculate the exact expected eth out amount and use that to ensure that the refund is correct

aleone (changes requested):
suggestion on how to do slippage control

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -0,0 +1,154 @@
...
+        uint256 flashSwappedCollateral;
+        uint256 totalCollateralToDeposit;
+        uint256 wPowerPerpAmount;
+    }
+
+    event FlashswapWMint(
+        address indexed depositor,
+        uint256 vaultId,
+        uint256 wPowerPerpAmount,
+        uint256 swapedCollateralAmount,
```
alpinechicken, line 51: *swappedCollateralAmount?

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -0,0 +1,154 @@
...
+            uint256 vaultId = IController(controller).mintWPowerPerpAmount{value: data.totalCollateralToDeposit}(
+                data.vaultId,
+                data.wPowerPerpAmount,
+                0
+            );
+
+            //repay the flash swap
+            IWPowerPerp(wPowerPerp).transfer(wPowerPerpPool, _amountToPay);
+
+            if (address(this).balance > 0) {
```
aleone, line 146: IMO this needs to be checking weth balance as we convert all to weth and then might pay less than the msg.value

aleone (changes requested):
we need to refund unused WETH to the user

if the slippage is not used they should get a refund

test is not testing right thing

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -3,30 +3,152 @@
...
```
aleone, line 146: what happens if we try to send 0 ETH?

for pretty much every case there will be some remaining eth in the contract (at least 1 wei), so is it cheaper to just send address(this).balance to the user ever time and not check, or will it revert?

aleone (approved):

aleone:
lgtm, but do think we should check that the user gets the expected eth amount back in tests (maybe you wanted to do in a test PR improvement later or something)

packages/hardhat/test/integration-tests/periphery/controller-helper.ts
```diff
@@ -0,0 +1,150 @@
...
+        tokenIn: wSqueeth.address,
+        tokenOut: weth.address,
+        fee: 3000,
+        recipient: owner.address,
+        deadline: Math.floor(await getNow(ethers.provider) + 8640000),
+        amountIn: mintWSqueethAmount,
+        amountOutMinimum: 0,
+        sqrtPriceLimitX96: 0
+      }    
+      await wSqueeth.connect(owner).approve(swapRouter.address, constants.MaxUint256)
```
KMKoushik, line 126: Doubt. Why do we need this? Is it to get `exactInputSingle`?

PR #175, opened 2022-02-17
feat: mint and LP

haythem96:
# feat: mint and LP

## Description

This PR add the feature to mint a short position and LP in Uniswap pool v3

Fixes # (issue)

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist


packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -162,12 +177,66 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
+        INonfungiblePositionManager.MintParams memory params = INonfungiblePositionManager.MintParams({
+            token0: token0,
+            token1: token1,
+            fee: IUniswapV3Pool(wPowerPerpPool).fee(),
+            tickLower: _lowerTick,
+            tickUpper: _upperTick,
+            amount0Desired: token0 == wPowerPerp ? _wPowerPerpAmount : _collateralToLP,
+            amount1Desired: token1 == wPowerPerp ? _wPowerPerpAmount : _collateralToLP,
+            amount0Min: 0,
+            amount1Min: 0,
```
haythem96, line 228: Any thoughts what those 2 should be ?

aleone: I think these should be set as a slippage param on f/e, so probably passed in.

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -162,12 +177,66 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
+        int24 _upperTick
+    ) external payable {
+        require(msg.value == _collateralToMint.add(_collateralToLP), "Wrong ETH sent");
+
+        uint256 vaultId = IController(controller).mintWPowerPerpAmount{value: _collateralToMint}(
+            _vaultId,
+            _wPowerPerpAmount,
+            0
+        );
+        address token0 = IUniswapV3Pool(wPowerPerpPool).token0();
```
aleone, line 216: cheaper to store on contract creation vs reading every time

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -162,12 +177,66 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
+        uint256 _vaultId,
+        uint256 _wPowerPerpAmount,
+        uint256 _collateralToMint,
+        uint256 _collateralToLP,
+        int24 _lowerTick,
+        int24 _upperTick
+    ) external payable {
+        require(msg.value == _collateralToMint.add(_collateralToLP), "Wrong ETH sent");
+
+        uint256 vaultId = IController(controller).mintWPowerPerpAmount{value: _collateralToMint}(
```
aleone, line 211: is it cheaper to declare a repeatedly used interface vs calling it in a function?

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -162,12 +177,66 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
+            token1: token1,
+            fee: IUniswapV3Pool(wPowerPerpPool).fee(),
+            tickLower: _lowerTick,
+            tickUpper: _upperTick,
+            amount0Desired: token0 == wPowerPerp ? _wPowerPerpAmount : _collateralToLP,
+            amount1Desired: token1 == wPowerPerp ? _wPowerPerpAmount : _collateralToLP,
+            amount0Min: 0,
+            amount1Min: 0,
+            recipient: msg.sender,
+            deadline: block.timestamp
```
aleone, line 230: uniswap lets users set this, but IMO its kinda worthless for LPing, I don't really care when my tx is mined

to match uniswap implementation it should be a param that is passed in

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -162,12 +177,66 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
+        });
+
+        INonfungiblePositionManager(nonfungiblePositionManager).mint{value: _collateralToLP}(params);
+
+        if (_vaultId == 0) IShortPowerPerp(shortPowerPerp).safeTransferFrom(address(this), msg.sender, vaultId);
```
aleone, line 235: should we let people deposit their NFT as collateral here, or do we want to have a separate function for it?

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -162,12 +177,66 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
+            tickUpper: _upperTick,
+            amount0Desired: token0 == wPowerPerp ? _wPowerPerpAmount : _collateralToLP,
+            amount1Desired: token1 == wPowerPerp ? _wPowerPerpAmount : _collateralToLP,
+            amount0Min: 0,
+            amount1Min: 0,
+            recipient: msg.sender,
+            deadline: block.timestamp
+        });
+
+        INonfungiblePositionManager(nonfungiblePositionManager).mint{value: _collateralToLP}(params);
```
aleone, line 233: don't we need to transfer the NFT to a user?

packages/hardhat/test/integration-tests/periphery/controller-helper.ts
```diff
@@ -154,4 +154,35 @@ describe("Controller helper integration test", function () {
...
+      const tokenIndexAfter = await (positionManager as INonfungiblePositionManager).totalSupply();
+      const tokenId = await (positionManager as INonfungiblePositionManager).tokenByIndex(tokenIndexAfter.sub(1));
+      const ownerOfUniNFT = await (positionManager as INonfungiblePositionManager).ownerOf(tokenId); 
+
+      expect(ownerOfUniNFT === depositor.address).to.be.true
+      expect(tokenIndexAfter.sub(tokenIndexBefore).eq(BigNumber.from(1))).to.be.true
+      expect(vaultBefore.shortAmount.eq(BigNumber.from(0))).to.be.true
+      expect(vaultBefore.collateralAmount.eq(BigNumber.from(0))).to.be.true
+      expect(vaultAfter.shortAmount.eq(mintWSqueethAmount)).to.be.true
+      expect(vaultAfter.collateralAmount.eq(collateralAmount)).to.be.true
```
aleone, line 185: probably should test some of the expected uniswap things?

probably should check if the transfered eth and osqth matches what we expect?

I think we need to refund users any excess oSQTH/ETH that isn't used in the uniswap NFT so need to check that

Maybe check the ticks or something?

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -162,12 +177,66 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
+
+        if (_vaultId == 0) IShortPowerPerp(shortPowerPerp).safeTransferFrom(address(this), msg.sender, vaultId);
+
+        emit BatchMintLp(msg.sender, _vaultId, _wPowerPerpAmount, _collateralToMint, _collateralToLP);
```
aleone, line 237: Think we need to return unused ETH and oSQTH to the user and write tests to ensure that this is as expected.

aleone (changes requested):
Some comments qs in the PR, one here:

How do you want to implement the flashloan LP case?

Could create a wrapper for this function that

borrows ETH via flashloan
calls batchMintLp
withdraws borrowed ETH from vault
repays ETH flashloan

packages/hardhat/contracts/periphery/ControllerHelper.sol
haythem96: idk, what u think ?

packages/hardhat/contracts/periphery/ControllerHelper.sol
haythem96: Recipient is user address, so should be transferred there directly

alpinechicken (approved):
lgtm - is this a special case of #191? Is it ever better to mint and LP without a flashswap?

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -143,30 +159,94 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
     function flashswapWBurnBuyLong(
         uint256 _vaultId,
         uint256 _wPowerPerpAmountToBurn,
         uint256 _wPowerPerpAmountToBuy,
         uint256 _collateralToWithdraw,
-        uint256 _maxToPay
+        uint256 _collateralToBuyWith
     ) external payable {
-        require(_maxToPay <= _collateralToWithdraw.add(msg.value), "Not enough collateral");
+        require(_collateralToBuyWith <= _collateralToWithdraw.add(msg.value), "Not enough collateral");
```
aleone, line 171: not sure why we made this change (do you know? did we discuss?), but think the require is not correct now at least if it means what I'm reading it to mean.

right now if the 5th param _collateralToBuyWith is the amount of vault collateral to use for buying oSQTH the require needs to be changed to be:

```
        require(_collateralToBuyWith <= _collateralToWithdraw, "Not enough collateral");

```

this is because you add msg.value to it later so it can use collateralToBuyWith + msg.value but still can't use more collateral than withdrawn

but honestly I think its more confusing this way - why did we change?

packages/hardhat/contracts/periphery/ControllerHelper.sol
aleone: see this? am I correct, shouldn't we store this on contract creation in constructor vs calling it externally for every use?

packages/hardhat/contracts/periphery/ControllerHelper.sol
aleone: kk

packages/hardhat/contracts/periphery/ControllerHelper.sol
aleone: yip, missed that

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -143,30 +159,94 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
+        uint256 _wPowerPerpAmount,
+        uint256 _collateralToMint,
+        uint256 _collateralToLP,
+        uint256 _amount0Min,
+        uint256 _amount1Min,
+        uint256 _deadline,
+        int24 _lowerTick,
+        int24 _upperTick
+    ) external payable {
+        require(msg.value == _collateralToMint.add(_collateralToLP), "Wrong ETH sent");
```
aleone, line 212: I think this needs to take into account _amount0Min or _amount1Min for extra ETH amount that might be needed.

When LPing the asset composition may change slightly from when tx sent and mined, so you might need a bit extra ETH or oSQTH (extra oSQTH is a problem, see below).

I also think that we need to ensure that the LP is always created with no more than _wPowerPerpAmount (as we mint a fixed amount), can we do this with the _amount0Min or _amount1Min?

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -143,30 +159,94 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
+        });
+
+        INonfungiblePositionManager(nonfungiblePositionManager).mint{value: _collateralToLP}(params);
+
+        if (_vaultId == 0) IShortPowerPerp(shortPowerPerp).safeTransferFrom(address(this), msg.sender, vaultId);
+        if (address(this).balance > 0) {
+            payable(msg.sender).sendValue(address(this).balance);
+        }
+        uint256 remainingWPowerPerp = IWPowerPerp(wPowerPerp).balanceOf(address(this));
+        if (remainingWPowerPerp > 0) {
```
aleone, line 243: should we just burn it vs returning it?

aleone: seems more annoying to get some back and have to manually burn it or sell it or something

packages/hardhat/contracts/periphery/ControllerHelper.sol
aleone: The alternative is that we mint extra oSQTH and then burn when unused.

aleone (commented):
have qs on how we handle the need for extra ETH as well as minting a fixed oSQTH amount and the potential for LP composition changes btwn tx submission and mining

haythem96:
Re this https://github.com/opynfinance/squeeth-monorepo/pull/175#discussion_r819839911
I think I did the improvement in another PR later, if not I will do in separate PR with bunch of opt

packages/hardhat/contracts/periphery/ControllerHelper.sol
haythem96: I think this is changed in a rebase

haythem96: remove this change

packages/hardhat/contracts/periphery/ControllerHelper.sol
haythem96: sounds good, will burn it

haythem96: But not sure if it is worth to do it here as I need to get the amount of collateral to withdraw, and require calling oracle..etc which maybe not that efficient, what u think ?

haythem96: Thought on just burning the debt without withdrawing collateral ?

packages/hardhat/contracts/periphery/ControllerHelper.sol
haythem96: Not sure I get ur comment, _amount0Min and _amount1Min used to check for slippage, making sure amounts deposited in LP are greater or equal.

`require(amount0 >= params.amount0Min && amount1 >= params.amount1Min, 'Price slippage check')`

aleone (changes requested):
some comments, 1 around refundETH()

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -167,6 +182,66 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
+            amount0Desired: token0 == wPowerPerp ? _wPowerPerpAmount : _collateralToLP,
+            amount1Desired: token1 == wPowerPerp ? _wPowerPerpAmount : _collateralToLP,
+            amount0Min: _amount0Min,
+            amount1Min: _amount1Min,
+            recipient: msg.sender,
+            deadline: _deadline
+        });
+
+        INonfungiblePositionManager(nonfungiblePositionManager).mint{value: _collateralToLP}(params);
+
```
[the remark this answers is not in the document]
haythem96: great catch!

aleone (approved):
lgtm

Issue #177, opened 2022-02-20
Short PnL loading issues


alexisgauba:
from user: 

for shorts - maybe its just the laggy thing, but seems to be worse than just lag
this AM I was up 150% repeatedly (loads and wait and never changed)

new load now:
briefly before further refreshing it showed I was up 0.5%
then showing down 80%
refreshed itself automatically and shows down 30%
now showing 0.5% again
seems to refresh itself a lot and sometimes hang maybe in mid calculation or something?

PR #170, opened 2022-02-15
Use react-query for API calls 

EstherSeyi:
# Task:

## Description

-  update API calls to use react-query

> What are the places that need testing?
- getETHPriceCoingecko - fetches current ethPrice from coingecko
- useETHPriceCharts

> What would be the improvement?
- The getETHPriceCoingecko fetch is not affected by re-renders, it’ll only fetch when it’s supposed to which is at the specified interval. You can test this by checking the network tab on the initial render to ensure that the ``https://api.coingecko.com/api/v3/simple/price?ids=${coin}&vs_currencies=usd`` network call occurs only once initially and updates only at the specified interval.
- The other cached API calls that do not run on intervals have a more consistent caching.

> The APIs that are now using react-query
- ``https://api.coingecko.com/api/v3/simple/price?ids=${coin}&vs_currencies=usd``
- `https://api.coingecko.com/api/v3/coins/ethereum/market_chart?vs_currency=usd&days=${day}`
- `https://api.coingecko.com/api/v3/coins/ethereum/market_chart?vs_currency=usd&days=90`
- `https://api.compound.finance/api/v2/market_history/graph?asset=0x39aa39c021dfbae8fac545936693ac917d5e7563&min_block_timestamp=${startTime}&max_block_timestamp=${endTime}&num_buckets=${day}`

I also moved two contract interactions that were being done at intervals(with useInterval hook) to also use react query for more consistent caching; `contract.methods.balanceOf` and `web3.eth.getBalance`

The other file changes are from updates due to typescript errors after the changes.

Fixes # (issue)
#169 
## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Document update

## How Has This Been Tested

Please describe how to test to verify the changes. Provide instructions so we can reproduce.

## Checklist

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] Added video recordings if it is a UI change

PR #181, opened 2022-02-21
Fix Pnl loading issues

EstherSeyi:
# Task:

## Description

> I found that the issue was because sometimes the historic price(twelvedata) data is not being fetched cos the number of calls exceeds our limit per minute. I was getting `{"code":429,"message":"You have run out of API credits for the current minute. 663 API credits were used, with the current limit being 377. Wait for the next minute or consider switching to a higher tier plan at https://twelvedata.com/pricing","status":"error”}` error and historic price was returning zero when that happened.

> What I have done is ;

- Reduce the number of calls being made per user by ensuring that we only fetch the historic price for swaps that are relevant to the current user's position (see getRelevantSwaps lib/pnl.ts), and cache fetched historic price so that it doesn't keep calling the endpoint for every re-render.
- Invalidate the pnl calculation if historic ethPrice is not fetched, so it will not show any pnl if it’s unable to fetch historic ethPrice. This will ensure that we are not giving the user the wrong information.

Fixes # (issue)
#177 
## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/packages/frontend/src/context/wallet.tsx
+++ b/packages/frontend/src/context/wallet.tsx
@@ -10,6 +10,7 @@ import { EtherscanPrefix } from '../constants'
 import useInterval from '@hooks/useInterval'
 import { Networks } from '../types'
 import { Web3Provider } from '@ethersproject/providers'
+import { useQueryClient } from 'react-query'
 
 const useAlchemy = process.env.NEXT_PUBLIC_USE_ALCHEMY
 const defaultWeb3 = useAlchemy
@@ -52,6 +53,8 @@ const WalletProvider: React.FC = ({ children }) => {
   const [signer, setSigner] = useState<any>(null)
   const [balance, setBalance] = useState<BigNumber>(new BigNumber(0))
 
+  const queryClient = useQueryClient()
+
   const onWalletSelect = useCallback(async () => {
     if (!onboard) return
     onboard.walletSelect().then((success) => {
@@ -62,6 +65,7 @@ const WalletProvider: React.FC = ({ children }) => {
   const disconnectWallet = useCallback(async () => {
     if (!onboard) return
     await onboard.walletReset()
+    queryClient.removeQueries()
     setAddress(null)
     setBalance(new BigNumber(0))
   }, [onboard])
--- a/packages/frontend/src/lib/pnl.ts
+++ b/packages/frontend/src/lib/pnl.ts
@@ -63,12 +63,15 @@ export async function calcETHCollateralPnl(
   uniswapEthPrice: BigNumber,
   currentVaultEthBalance: BigNumber,
 ) {
-  const { deposits, withdrawals } = data?.length
+  const { deposits, withdrawals, priceError } = data?.length
     ? await data?.reduce(async (prevPromise, vaultHistory: VaultHistory_vaultHistories) => {
         const acc = await prevPromise
 
         const ethPriceAtTimeOfAction = await getEthPriceAtTransactionTime(vaultHistory.timestamp)
 
+        if (ethPriceAtTimeOfAction?.isZero()) {
+          acc.priceError = true
+        }
         if (vaultHistory.action === Action.DEPOSIT_COLLAT) {
           acc.deposits = acc.deposits.plus(
             new BigNumber(toTokenAmount(vaultHistory.ethCollateralAmount, 18)).times(ethPriceAtTimeOfAction),
@@ -79,8 +82,11 @@ export async function calcETHCollateralPnl(
           )
         }
         return acc
-      }, Promise.resolve({ deposits: BIG_ZERO, withdrawals: BIG_ZERO }))
-    : { deposits: BIG_ZERO, withdrawals: BIG_ZERO }
+      }, Promise.resolve({ deposits: BIG_ZERO, withdrawals: BIG_ZERO, priceError: false }))
+    : { deposits: BIG_ZERO, withdrawals: BIG_ZERO, priceError: true }
+
+  return !priceError ? currentVaultEthBalance.times(uniswapEthPrice).minus(deposits.minus(withdrawals)) : BIG_ZERO
+}
 
 const getRelevantSwaps = (squeethAmount: BigNumber, swaps: swaps_swaps[], isWethToken0: boolean, isLong = false) => {
   let totalSqueeth = BIG_ZERO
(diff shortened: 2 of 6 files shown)


packages/frontend/src/lib/pnl.ts
```diff
@@ -70,25 +82,47 @@ export async function calcETHCollateralPnl(
...
-      }, Promise.resolve({ deposits: BIG_ZERO, withdrawals: BIG_ZERO }))
-    : { deposits: BIG_ZERO, withdrawals: BIG_ZERO }
+      }, Promise.resolve({ deposits: BIG_ZERO, withdrawals: BIG_ZERO, priceError: false }))
+    : { deposits: BIG_ZERO, withdrawals: BIG_ZERO, priceError: true }
 
-  return currentVaultEthBalance.times(uniswapEthPrice).minus(deposits.minus(withdrawals))
+  return !priceError ? currentVaultEthBalance.times(uniswapEthPrice).minus(deposits.minus(withdrawals)) : BIG_ZERO
+}
+
+const getRelevantSwaps = (squeethAmount: BigNumber, swaps: swaps_swaps[], isWethToken0: boolean, isLong = false) => {
```
yichiehliu, line 91: Do we make it with useMemo?

packages/frontend/src/lib/pnl.ts
```diff
@@ -70,25 +82,47 @@ export async function calcETHCollateralPnl(
...
+  let totalSqueeth = BIG_ZERO
+  const relevantSwaps = []
+  for (let index = swaps.length - 1; index >= 0; index--) {
+    const squeethAmt = new BigNumber(isWethToken0 ? swaps[index].amount1 : swaps[index].amount0)
```
alexisgauba, line 95: this var name and the following are pretty confusing imo. would try to use smth that describes what each represents better

packages/frontend/src/lib/pnl.ts
alexisgauba, line 91: what is defined to be a 'relevant swap'? I personally don't understand the logic of what we're defining as relevant here? Also might be good to add this as a comment describing the function

alexisgauba:
Does #170 need to be merged before this one?

packages/frontend/src/lib/pnl.ts
```diff
@@ -70,25 +82,47 @@ export async function calcETHCollateralPnl(
...
+    const sqthAmount = isLong ? squeethAmt.negated() : squeethAmt
+    totalSqueeth = totalSqueeth.plus(sqthAmount)
+    relevantSwaps.push(swaps[index])
+    if (squeethAmount.isEqualTo(totalSqueeth)) {
```
yichiehliu, line 99: Just wondering if this really works cuz if we going from the end of swaps data, the summation of users' position might go wrong

alexisgauba:
It's just showing me 0, is this expected behavior for if the API hasn't loaded? Thinking we should probably show a loading message instead since 0 could mislead users to think they are neutral from when they opened their position 

Screen Shot 2022-02-21 at 2 31 53 PM

packages/frontend/src/lib/pnl.ts
EstherSeyi: It's asc so the most recent data are at the end.

EstherSeyi:
> Does #170 need to be merged before this one? 

No, anyone can be merged first

EstherSeyi:
> It's just showing me 0, is this expected behavior for if the API hasn't loaded? Thinking we should probably show a loading message instead since 0 could mislead users to think they are neutral from when they opened their position 
> 
> 
> 
> Screen Shot 2022-02-21 at 2 31 53 PM
> 
> 

Okay, will update.

packages/frontend/src/lib/pnl.ts
EstherSeyi: it's not in a hook or component

yichiehliu: Sorry my bad, I thought it was

alexisgauba:
Took around 5min to load PnL for me - any idea why it might be taking so long?

EstherSeyi:
> Took around 5min to load PnL for me - any idea why it might be taking so long?

Yes, it's probably because it was unable to fetch historic prices because of the number calls. I think merging this branch will reduce the number of calls and make it better.
Screenshot 2022-02-23 at 07 39 51

packages/frontend/src/lib/pnl.ts
alexisgauba: Hmm still don't fully understand this - what lines are specifically getting swaps related to a user's current position? Where is the logic that defines what swap is related to a users current position vs. not?

EstherSeyi: Starting from the most recent swap.
Line 105 - is adding squeeth amount of current swap to a variable "totalSqueeth"
Line 106 - is pushing the swap into the relevantSwaps array
Line 109 - is checking if the accumulated squeeth adds up to the user position squeeth amount.
Line 110 - is stoping the loop if condition on line 109 is true.
Line 113 - is returning the array with the swaps that got pushed into the array

alexisgauba: Ok I guess I don't understand how this is different than just using the swaps array? Seems like we are taking the swaps array, pushing all swaps into the relevant swaps array(?), and then what is the user position squeeth amount, like where is that coming from and how is that different than the accumulated squeeth here?

yichiehliu: Not sure if this will work as expected, i.e. if users buy and then mint oSQTH, isLong will end up being true, and will cause the squeethAmt of mint position seems to be added up as totalSqueeth in line 96&97

EstherSeyi: @yichiehliu minted squeeth is not included in the swaps or squeethAmount. Minted squeeth is calculated `oSqueethBal.minus(squeethAmount)` in the code.
@alexisgauba Okay the user's squeeth is passed in as an argument from the positions context. Some of the swaps in the swaps array may be from previous positions that have been completely closed, so this ensures that we don't include them.

yichiehliu: @EstherSeyi Actually it is, cuz we are getting all the swaps data from a user, which includes mint/burn txs

yichiehliu: Ok, I think I was just confused myself with direct selling on uniswap

packages/frontend/src/lib/pnl.ts
yichiehliu: Let's say user has long 2 osqth first, and then mint 2 osqth and sell 2 osqth, from the end of swaps, you will get -2(sell), +2(long) = 0, but the actual position will be 2 long position and 2 short position, so if we consider direct selling, which we already did rn, the calculation of actual long & short will not be just accumulated all txs up, so prob we cannot take swaps data like this, and Im still working on to get correct direct selling short position 

This will work for most users who did not do direct selling on uniswap, so not sure if we want to have this temporary fix here

alexisgauba: Ok so @yichiehliu you'd fix this in the next PR ?

yichiehliu: yh so I think can merge this one first and then update the pnl calcs then

yichiehliu (approved):

packages/frontend/src/context/wallet.tsx
```diff
@@ -71,6 +70,7 @@ const WalletProvider: React.FC = ({ children }) => {
   const disconnectWallet = useCallback(async () => {
     if (!onboard) return
     await onboard.walletReset()
+    queryClient.removeQueries()
```
KMKoushik, line 73: Niceee!

KMKoushik (approved):

PR #196, opened 2022-02-24
use squeeth atoms

ezhou7:
# Task:

Users complaining of lag when loading web pages.

## Description

I believe the issue is arising b/c `useSqueethPool()` resets the atoms when it is called, so users see slow loading. There is still some lag from what I can tell, so this is by no means a permanent fix.

Fixes # (issue)

## Type of change

- [ ] Document update

## How Has This Been Tested

Tested with `yarn dev` and `yarn build && yarn start`. Since I've only tested locally with incomplete access to Blocknative and Firestore, my local testing may not line up with actual UX experience, so please test on your end as well with the preview. Thanks!

## Checklist

### DIFF
--- a/packages/frontend/pages/positions.tsx
+++ b/packages/frontend/pages/positions.tsx
@@ -154,7 +154,7 @@ export function Positions() {
     longUnrealizedPNL,
   } = usePnL()
 
-  const pool = useAtom(poolAtom)[0]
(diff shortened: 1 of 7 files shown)


akwritescode:
discussed with @ezhou7 in texts and he agrees that we shud revert now and he will re-introduce in a new PR that is performing much better

PR #197, opened 2022-02-25
feat: sell long and flash mint short

haythem96:
# feat: sell long and flash mint short

## Description

This PR add the feature to sell long wPowerPerp, mint short position using the ETH from the first swap, flashswap and some amount sent from user, and sell short position to repay flashswap.

- Sell `X` oSQTH for `Y` ETH, set `minToReceive`
- Get `A` vault collateral, flashswap `A-msg.value-Y`
- Mint `Z` oSQTH, and repay flashswap

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist


packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -189,6 +197,47 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
+    /**
+     * @notice sell long wPowerPerp and flashswap mint short position
+     * @dev flahswap amount = collateral amount - msg.value - ETH from selling long wPowerPerp
+     * @param _vaultId vault ID
+     * @param _wPowerPerpAmountToMint wPowerPerp amount to mint
+     * @param _collateralAmount collateral amount to use for minting
+     * @param _wPowerPerpAmountToSell long wPowerPerp amount to sell
+     * @param _minToReceive min ETH amount to receive for selling long wPowerPerp amount
+     */
+    function flashswapSellLongWMint(
```
aleone, line 209: this can be simplified/cheaper as 1 exactInFlashSwap right (similar to the sell long, mint + sell)

- do an exactInFlashSwap for the _wPowerPerpAmountToSell + _wPowerPerpAmountToMint (this requires slippage in terms of both quantities, so the last param is the minToReceive for the total sell)
- mint _wPowerPerpAmount using _collateralAmount (which can be a combo of msg.value and proceeds from step 1)
- repay #1 using the minted amount and the transferred in oSQTH

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -189,6 +197,47 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
+            ""
+        );
+
+        // flahswap and mint short position
+        _exactInFlashSwap(
+            wPowerPerp,
+            weth,
+            IUniswapV3Pool(wPowerPerpPool).fee(),
+            _wPowerPerpAmountToMint,
+            _collateralAmount.sub(msg.value).sub(amountOut),
```
aleone, line 235: don't think this is correct/works

ex: if the user is selling a lot of oSQTH (which gives a large amountOut), and then only shorting a small amount, it could revert

think the above refactor would make this more clear

aleone: a user may be due a refund from this tx (they sell $1000 long oSQTH and then short $1 more w $2 of collateral)

aleone (changes requested):
imo should refactor to do 1 flash swap instead of 2, as well as the limit price for the 2nd flashswap has issuse

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -189,6 +197,47 @@ contract ControllerHelper is FlashControllerHelper, IERC721Receiver {
...
+     * @notice sell long wPowerPerp and flashswap mint short position
+     * @dev flahswap amount = collateral amount - msg.value - ETH from selling long wPowerPerp
+     * @param _vaultId vault ID
+     * @param _wPowerPerpAmountToMint wPowerPerp amount to mint
+     * @param _collateralAmount collateral amount to use for minting
+     * @param _wPowerPerpAmountToSell long wPowerPerp amount to sell
+     * @param _minToReceive min ETH amount to receive for selling long wPowerPerp amount
+     */
+    function flashswapSellLongWMint(
+        uint256 _vaultId,
```
alpinechicken, line 210: Would it make sense to include existing vault collateral too to _collateralAmount is final amount in the vault?

haythem96: What we need the final vault collateral amount for ? `_collateralAmount` is for the amount to use for minting.

packages/hardhat/contracts/periphery/ControllerHelper.sol
aleone: looks good!

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -3,30 +3,365 @@
...
+    }
+
+    /**
+     * @notice sell long wPowerPerp and flashswap mint short position
+     * @dev flahswap amount = collateral amount - msg.value - ETH from selling long wPowerPerp
+     * @param _vaultId vault ID
+     * @param _wPowerPerpAmountToMint wPowerPerp amount to mint
+     * @param _collateralAmount collateral amount to use for minting
+     * @param _wPowerPerpAmountToSell long wPowerPerp amount to sell
+     * @param _minToReceive min ETH amount to receive for selling long wPowerPerp amount
```
aleone, line 207: this now should be updated to be min ETH amount to receive for selling _wPowerPerpAmountToSell and _wPowerPerpAmountToMint combined

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -3,30 +3,365 @@
...
+     */
+    function flashswapSellLongWMint(
+        uint256 _vaultId,
+        uint256 _wPowerPerpAmountToMint,
+        uint256 _collateralAmount,
+        uint256 _wPowerPerpAmountToSell,
+        uint256 _minToReceive
+    ) external payable {
+        IWPowerPerp(wPowerPerp).transferFrom(msg.sender, address(this), _wPowerPerpAmountToSell);
```
aleone, line 216: should we have a require here (may just fail earlier if incorrectly sent tx vs waiting to fail in the callback so probably not strictly necessary)?

```require(_minToReceive.add(msg.value)>=_collateralAmount)```

packages/hardhat/contracts/periphery/ControllerHelper.sol
aleone: are you thinking that someone may want to sell and withdraw collateral from the vault at same time?  is that the use case you are thinking?

maybe I should list every user flow today and make sure we check them all off

gamma's operate() functionality is nice for these kinds of things

aleone: or thinking that someone wants to sell long, mint short sell, and withdraw collateral

aleone (approved):
think the rebase is messed up - reviewed the changes from last commit and the flashswapSellLongWMint() function

small comments

Issue #205, opened 2022-03-02
Page content scrolls into navbar


rootulp:
**Describe the bug**
Page content scrolls into navbar

**To Reproduce**

1. Go to https://squeeth.opyn.co/
1. Scroll down

**Expected behavior**
A. Navbar and page content both scroll down or
B. If page has a static navbar, content scrolls behind navbar

**Screenshots**
<img width="707" alt="Screen Shot 2022-03-01 at 9 43 33 PM" src="https://user-images.githubusercontent.com/3699047/156284946-76c2a1d6-e928-4c6f-824e-1eec6b6a704d.png">

**Additional context**
It may be worth exploring `display: fixed` [here](https://github.com/opynfinance/squeeth-monorepo/blob/main/packages/frontend/src/components/Nav.tsx#L26) or consider using [`<AppBar>`](https://mui.com/components/app-bar/)

KMKoushik:
@rootulp Thanks for raising this issue. It works fine for the chrome browser. I think the issue is with the [backdrop filter](https://github.com/opynfinance/squeeth-monorepo/blob/main/packages/frontend/src/components/Nav.tsx#L31). Doc mentions that backdrop filter is not available for [Mozilla browser](https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter#browser_compatibility).

Will fix it. Appreciate any suggestions!

<img width="1440" alt="Screenshot 2022-03-02 at 3 30 36 PM" src="https://user-images.githubusercontent.com/24666922/156295277-50a5f0d2-812c-4e91-b300-639a2f09976d.png">

rootulp:
You're right! I can reproduce this on Firefox but can't reproduce on Chrome / Brave.

PR #207, opened 2022-03-02
fix(nav): prevent content scrolling into nav

rootulp:
Fixes https://github.com/opynfinance/squeeth-monorepo/issues/205

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Screenshot

Before | After
--- | ---
Screen Shot 2022-03-02 at 12 33 37 AM | Screen Shot 2022-03-02 at 12 31 06 AM

### DIFF
--- a/packages/frontend/src/components/Nav.tsx
+++ b/packages/frontend/src/components/Nav.tsx
@@ -28,7 +28,7 @@ const useStyles = makeStyles((theme) =>
       zIndex: 30,
       //background: theme.palette.background.default,
       borderBottom: `1px solid ${theme.palette.background.stone}`,
-      backdropFilter: 'blur(30px)',
+      backgroundColor: 'black',
     },
     logo: {
       marginRight: 'auto',


packages/frontend/src/components/Nav.tsx
```diff
@@ -28,7 +28,7 @@ const useStyles = makeStyles((theme) =>
       zIndex: 30,
       //background: theme.palette.background.default,
       borderBottom: `1px solid ${theme.palette.background.stone}`,
-      backdropFilter: 'blur(30px)',
+      backgroundColor: 'black',
```
KMKoushik, line 31: Thanks for the PR. Instead of removing the backdrop altogether, Can we apply different yet subtle background only for Mozilla? That way we retain the original style for other browsers!

https://stackoverflow.com/questions/56593433/how-to-write-mozilla-specific-css-using-withstyles

KMKoushik (changes requested):

packages/frontend/src/components/Nav.tsx
rootulp: We can. Per the [backdrop filter](https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter#browser_compatibility) page you linked, Mozzila and Internet Explorer don't support backdrop filter so I suspect we would want to special case styles for both those browsers.

I don't have visibility into the front-end designs for Squeeth so I don't know what styles this nav should conform to. I'll close this PR and let maintainers tackle https://github.com/onbjerg/foundry-book/pull/79 in whichever way aligns with front-end design styles.

PR #257, opened 2022-03-23
Flash Swap & Mint

EstherSeyi:
# Task:

## Description
See story [here](https://linear.app/opyn/issue/ENG-330/open-increase-short)
- Integrates Controller helper contract `flashswapSellLongWMint` method for the new open short feature.
- Updates subgraph to document and add FLASH_SWAP_W_MINT actions to vault history.
- Updates swaps to include controllerhelper address recipient.

Fixes # (issue)

## Type of change

- [ ] Document update

## How Has This Been Tested
Test by opening short position
use these to test [subgraph](https://thegraph.com/hosted-service/subgraph/estherseyi/sample)
HTTP: https://api.thegraph.com/subgraphs/name/estherseyi/sample
subscription: wss://api.thegraph.com/subgraphs/name/estherseyi/sample

## Checklist


alexisgauba:
Seems like this is still in progress? Will share some initial feedback as I was going through it: 

Looks like the liquidation price is wrong -> if its less than the ETH price this vault should have already been liquidated 
Screen Shot 2022-04-18 at 3 28 30 PM

Since the ControllerHelper takes the user's ETH premiums and uses them as collateral to mint more squeeth, the user shouldn't receive any ETH as "premium". Thinking we remove that "initial premium" line and change the top copy to something like "Mint and sell squeeth to receive funding". cc @aleone @alpinechicken to confirm re users not receiving premiums directly 
Screen Shot 2022-04-18 at 3 28 55 PM

Imo we should update this copy and this tooltip
Screen Shot 2022-04-18 at 6 50 15 PM

Not sure why this changed from "sell" to "minted sqth", users will still be selling. Also this is from a brand new fresh account with 0 squeeth, so not sure where it is getting the 0.0156 from (ropsten addr: `0xf1c0E71F49abB9DC3F95CdA43C23e18AC7BdF933`)
Screen Shot 2022-04-18 at 6 56 14 PM

When I clicked the "allow wrapper to manage vault" I'm getting a revert 
Screen Shot 2022-04-18 at 7 02 36 PM

packages/frontend/src/utils/useDebounce.tsx
```diff
@@ -0,0 +1,22 @@
+import { useState, useEffect } from 'react'
+
+export default function useDebounce(value: any, delay: number) {
```
adam5909099, line 3: This hook looks cool. I think it'll only work when the value is primitive. Might be better if we use explicit typing string | number | boolean.

adam5909099 (approved):
Looks good to me.

packages/frontend/src/components/Trade/Short/OpenShort.tsx
```diff
@@ -0,0 +1,732 @@
...
+      marginRight: 'auto',
+    },
+  }),
+)
+
+const FUNDING_MOVE_THRESHOLD = 0.7
+
+export const OpenShortPosition = () => {
+  const classes = useStyles()
+  const [confirmedAmount, setConfirmedAmount] = useState('0')
```
KMKoushik, line 134: I am concerned with the amount states variables we are using. The trade page flickers a lot

KMKoushik: How can we make this better?

packages/frontend/src/components/Trade/Short/OpenShort.tsx
```diff
@@ -0,0 +1,732 @@
...
+
+  const handleSqthChange = useAppCallback(
+    (val: string, collatPercent: number) => {
+      onSqthChange(val, collatPercent)
+    },
+    [onSqthChange],
+  )
+
+  useAppEffect(() => {
+    if (!debouncedAmount || debouncedAmount.lte(0)) return setExactOutAmount(BIG_ZERO)
```
KMKoushik, line 285: Too many app effects, Can you add a comment on what each one does?

KMKoushik (changes requested):
Looks fine, but it's very hard to actually change the collat ratio. It keeps on changing

packages/frontend/src/components/Trade/Short/OpenShort.tsx
EstherSeyi: Hey @KMKoushik, I'm not sure what you mean when you say it's very hard to change the collat ratio. Currently, there is a minimum collat ratio value, below which the user cannot set their collat ratio.

EstherSeyi: @KMKoushik about the state variables, I have removed three of them that I thought were redundant.

adam5909099: I made PR up for fixing the flickering issue @KMKoushik mentioned. After playing his video slowly, I realized that flickering is related to the vault that is sometimes defined but not defined sometimes on his local. Its changes are too fast, so we see flickering. I can't reproduce the same thing on my local but I think it's related to a performance issue in the vault manager.
https://github.com/opynfinance/squeeth-monorepo/pull/560

packages/frontend/src/utils/useDebounce.tsx
```diff
@@ -0,0 +1,23 @@
+import BigNumber from 'bignumber.js'
+import { useState, useEffect } from 'react'
+
+export default function useDebounce<T extends Number | BigNumber | String | Boolean>(value: T, delay: number) {
+  // State and setters for debounced value
+  const [debouncedValue, setDebouncedValue] = useState(value)
+  useEffect(
```
adam5909099, line 7: I can see value can be a big number in the function definition. I think we should use useAppEffect.

alexisgauba:
Here my collRatio is changing but my liquidation price isn't - why is that the case? 

https://user-images.githubusercontent.com/13340486/177686492-8322bc12-82e8-48a7-9495-f3a859c50b22.mov

This vault only has 14 minted oSQTH to sell, but I am trying to sell far more than that. How come it's not telling me I need more collateral? 

Screen Shot 2022-07-06 at 4 08 11 PM
https://user-images.githubusercontent.com/13340486/177686532-8cb5aba5-0908-4a8c-8bb7-de8a3b402497.mov

Why is it blocking me from reducing collateral? Also if it needs to do this block for some reason, we should have some explanation of why it is doing it so the user is not confused

https://user-images.githubusercontent.com/13340486/177686599-1d643147-d3fe-4980-b880-b0df498bdb80.mov

PR #260, opened 2022-03-24
Update depdancy array

### LINKED ISSUE #256: Crab PNL is not loading properly
**Describe the bug**
Screenshot 2022-03-23 at 7 45 20 PM

**To Reproduce**

1. Go to https://squeeth.opyn.co/positions with crab position
2. Refresh to see wrong crab position

adam5909099:
# Task:
Crab PNL is not loading properly

## Description
Update memo and useEffect dependency

Fixes #256 
https://github.com/opynfinance/squeeth-monorepo/issues/256

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/packages/frontend/src/hooks/useCrabPosition.ts
+++ b/packages/frontend/src/hooks/useCrabPosition.ts
@@ -25,6 +25,10 @@ export const useCrabPosition = (user: string) => {
   const [minPnlUsd, setMinPnlUsd] = useState(BIG_ZERO)
   const [minPnL, setMinPnL] = useState(BIG_ZERO)
 
+  const depositedDepedancy = data
+    ?.map((item) => `${item.ethAmount.toString()}:${item.lpAmount.toString()}:${item.ethUsdValue.toString()}`)
+    .join(' ')
+
   const { depositedEth, usdAmount: depositedUsd } = useMemo(() => {
     if (loading || !data) return { depositedEth: BIG_ZERO, usdAmount: BIG_ZERO }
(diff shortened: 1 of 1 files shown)


packages/frontend/src/hooks/useCrabPosition.ts
```diff
@@ -25,6 +25,10 @@ export const useCrabPosition = (user: string) => {
   const [minPnlUsd, setMinPnlUsd] = useState(BIG_ZERO)
   const [minPnL, setMinPnL] = useState(BIG_ZERO)
 
+  const depositedDepedancy = data
+    ?.map((item) => `${item.ethAmount.toString()}:${item.lpAmount.toString()}:${item.ethUsdValue.toString()}`)
+    .join(' ')
```
EstherSeyi, line 30: can we use data.length as the dependency instead or is this a better dependency?

adam5909099: The issue is happening because the calculation isn't handled even after the updated data. Are we sure that there's no case for having updated data (ethAmount or ethUsdValue) though there's no length change?

KMKoushik: This is actually a nice idea IMO! History will not change but I think this is the best idea for other cases like changing to a different address with the same length!

block902 (approved):
Looks good.

KMKoushik (approved):

Issue #258, opened 2022-03-23
Fetch and store contract states in _app.tsx


KMKoushik:
The contract states that we use often like `index`, `mark` currently being used as hooks and fetched every time we call. Though this will not affect performance it will make lot of API calls to RPC. 

Instead, we can fetch them and store them as atom on a global page like `_app.tsx` so we can use this atom directly everywhere

PR #261, opened 2022-03-24
Fetch and store contract states in _app.tsx

EstherSeyi:
# Task:

## Description

- Moves useIndex and useMark hooks to _app.tsx
- replaces useIndex and useMark hooks with atoms across the app.

Fixes # (issue)
#258 
## Type of change

- [ ] Document update

## How Has This Been Tested

- Tested by checking the mark and index have correct value compared with prod.

## Checklist


block902 (approved):
Looks good on my side.

yichiehliu (approved):
LGTM

packages/frontend/pages/_app.tsx
```diff
@@ -75,6 +76,8 @@ function MyApp({ Component, pageProps }: any) {
 const Init = () => {
   useUpdateSqueethPrices()
   useUpdateSqueethPoolData()
+  useIndex()
```
KMKoushik, line 79: Instead of importing each function 1 by 1, how about an init function in the controller hook, export that, and use it here?

```
export const useInitController = () => {
  useIndex()
  useMark()
  useCurrentImpliedFunding()
  useDailyHistoricalFunding()
  useNormFactor()
}
```

KMKoushik: Also we can remove export from function like useIndex(), useMark() so no one will accidentally call that instead of using atoms

KMKoushik (changes requested):

yichiehliu (commented):
Would it be more clear to users if we add "loading" to squeeth info instead of showing 0? 
cc @alexisgauba

alexisgauba:
> 

yes if its loading would be better to show a loading state

yichiehliu (approved):
LGTM

KMKoushik (approved):

PR #272, opened 2022-03-29
Custom hooks for handling bignumber depdancy

adam5909099:
# Task:

## Description
We used to have an eslint issue using a stringified version of Big number inside hooks dependency.
image
The big number can't be used as a dependency directly because it's an object, and lint shouldn't complain about a stringified big number. So, we tend to ignore dependency lint issues and sometimes miss necessary dependency though lint catches it, which causes critical UI bugs. 

This PR introduces useAppEffect, useAppMemo, useAppCallback, which you can pass Big number as a dependency directly without string conversion.
image
If you miss a necessary dependency, lint will complain about it. We shouldn't ignore any lint issues in new hooks.

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Document update

## How Has This Been Tested

Please describe how to test to verify the changes. Provide instructions so we can reproduce.

## Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] Added video recordings if it is a UI change

PR #290, opened 2022-04-05
Fix: crab UI show wrong PnL 

KMKoushik:
# Task:
Crab shows bloated values

Fixes # (issue)

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Document update

## How Has This Been Tested
Checked production values

## Checklist

### DIFF
--- a/packages/frontend/src/hooks/useCrabPosition.ts
+++ b/packages/frontend/src/hooks/useCrabPosition.ts
@@ -8,6 +8,7 @@ import { useAtomValue } from 'jotai'
 import { useTokenBalance } from './contracts/useTokenBalance'
 import { addressesAtom } from 'src/state/positions/atoms'
 import { indexAtom } from 'src/state/controller/atoms'
+import useAppCallback from './useAppCallback'
 
 export const useCrabPosition = (user: string) => {
   const crabLoading = useAtomValue(crabLoadingAtom)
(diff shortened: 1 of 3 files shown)


packages/frontend/src/hooks/useCrabPosition.ts
```diff
@@ -8,14 +8,14 @@ import { useAtomValue } from 'jotai'
 import { useTokenBalance } from './contracts/useTokenBalance'
 import { addressesAtom } from 'src/state/positions/atoms'
 import { indexAtom } from 'src/state/controller/atoms'
+import useAppCallback from './useAppCallback'
```
alexisgauba, line 11: what is useAppCallback? / how does it differ from normal useCallback?

alexisgauba (commented):
Been long enough since I've really gone deep into the code to have a good sense on the implementation here 

Looks ok to me when I test in prod, but what were the key issues and how did you solve them? 

Also is it worth adding a test for the fix per Anil's suggestion or does it only make sense to start doing that after the testing PR is merged?

packages/frontend/src/hooks/useCrabPosition.ts
KMKoushik: This is a new hook added by @adam5909099  https://github.com/opynfinance/squeeth-monorepo/pull/272 Helps to do the `.toString` of BigNumbers automatically so we don't have to ignore any warnings thereby missing any important dependency is missed.

KMKoushik: Also can't have tests until we find out how to test PnLs correctly

alexisgauba (approved):

adam5909099 (approved):
Looks great to me!

PR #312, opened 2022-04-07
fix: wrapping ETH for flashloan payment

### LINKED ISSUE #298: Some flashloans don't wrap back to ETH amt owed + premium
There needs to be a wrapping of weth (flashloan amount + premium) in the calling function or in the call back.  Right now the contract often will have no WETH unless sold some oSQTH for WETH on uniswap and the flashloan will fail.  Collateral withdrawn will be ETH, so won't be able to repay flashloan + premium.

Issue in following functions:
-flashloanCloseVaultLpNft

haythem96:
# fix: wrapping ETH for flashloan payment

## Description

Fixes #298 

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/packages/hardhat/contracts/periphery/ControllerHelper.sol
+++ b/packages/hardhat/contracts/periphery/ControllerHelper.sol
@@ -530,6 +530,11 @@ contract ControllerHelper is UniswapControllerHelper, AaveControllerHelper, IERC
                 data.collateralToWithdraw.add(data.collateralToFlashloan),
                 data.limitPriceEthPerPowerPerp
             );
+
+            if (address(this).balance > 0) {
+                // convert flashloaned amount + fee from ETH to WETH to prepare for payback
+                IWETH9(ControllerHelperDiamondStorage.getAddressAtSlot(5)).deposit{value: address(this).balance}();
+            }
         } else if (
             ControllerHelperDataType.CALLBACK_SOURCE(_callSource) ==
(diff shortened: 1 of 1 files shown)


packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -530,6 +530,11 @@ contract ControllerHelper is UniswapControllerHelper, AaveControllerHelper, IERC
                 data.collateralToWithdraw.add(data.collateralToFlashloan),
                 data.limitPriceEthPerPowerPerp
             );
+
```
alpinechicken, line 533: How does this weth get returned if nonzero?

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -530,6 +530,11 @@ contract ControllerHelper is UniswapControllerHelper, AaveControllerHelper, IERC
...
+            if (address(this).balance > 0) {
+                // convert flashloaned amount + fee from ETH to WETH to prepare for payback
+                IWETH9(ControllerHelperDiamondStorage.getAddressAtSlot(5)).deposit{value: address(this).balance}();
```
aleone, line 536: wonder if we should be consistent here across how we wrap to weth to repay the flashswap

there may be some edge cases we miss if we wrap amount+premium in some cases vs balance in other cases

wrapping the full balance is probably safer, as there could be cases where there is weth that can be used for the premium and eth used for the flashswap

aleone (approved):
lgtm

PR #322, opened 2022-04-08
Check vault ownership in parent function

### LINKED ISSUE #303: FLASHLOAN_REBALANCE_VAULT_NFT callback ownership check in parent function
Can we move the ownership check in the FLASHLOAN_REBALANCE_VAULT_NFT callback from where it is to the parent function to be consistent with the other ownership require statements.

haythem96:
# Check vault ownership in parent function

## Description

Fixes #303 

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/packages/hardhat/contracts/periphery/ControllerHelper.sol
+++ b/packages/hardhat/contracts/periphery/ControllerHelper.sol
@@ -407,6 +407,9 @@ contract ControllerHelper is UniswapControllerHelper, AaveControllerHelper, IERC
         uint256 _collateralToFlashloan,
         ControllerHelperDataType.RebalanceVaultNftParams[] calldata _params
     ) external payable {
+        // check ownership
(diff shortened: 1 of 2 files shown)


aleone (approved):
lgtm

aleone (changes requested):
Any ETH sent by calling rebalanceVaultNft is never wrapped to WETH, so trying to trade from ETH to oSQTH to rebalance via the general swap wont work. 

One potential solution is wrapping the ETH balance before the call, and then unwrapping after the callback.  I fear these functions are going to be mega gas intensive.

aleone (changes requested):
See comment on ETH to oSQTH swaps being limited as we don't wrap any ETH to WETH and need to repay in WETH.

aleone (changes requested):
See comment on wrapping weth

PR #325, opened 2022-04-10
feat: general swap

haythem96:
# feat: general swap

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/packages/hardhat/contracts/periphery/ControllerHelper.sol
+++ b/packages/hardhat/contracts/periphery/ControllerHelper.sol
@@ -808,7 +808,7 @@ contract ControllerHelper is UniswapControllerHelper, AaveControllerHelper, IERC
             ControllerHelperDataType.CALLBACK_SOURCE(_callSource) ==
             ControllerHelperDataType.CALLBACK_SOURCE.GENERAL_SWAP
         ) {
-            IWETH9(ControllerHelperDiamondStorage.getAddressAtSlot(5)).deposit{value: msg.value}();
+            IWETH9(ControllerHelperDiamondStorage.getAddressAtSlot(5)).deposit{value: address(this).balance}();
             IERC20(_tokenIn).transfer(ControllerHelperDiamondStorage.getAddressAtSlot(3), _amountToPay);
         }
     }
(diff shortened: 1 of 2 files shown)


alpinechicken (approved):
lgtm

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -786,6 +804,11 @@ contract ControllerHelper is UniswapControllerHelper, AaveControllerHelper, IERC
                 ControllerHelperDiamondStorage.getAddressAtSlot(3),
                 _amountToPay
             );
+        } else if (
+            ControllerHelperDataType.CALLBACK_SOURCE(_callSource) ==
+            ControllerHelperDataType.CALLBACK_SOURCE.GENERAL_SWAP
+        ) {
+            IERC20(_tokenIn).transfer(ControllerHelperDiamondStorage.getAddressAtSlot(3), _amountToPay);
```
aleone, line 811: this only works for WETH/ETH to oSQTH swaps when we have a WETH balance in the contract

any payable ETH can't be used for this as it isn't ever wrapped and any withdrawn ETH from a vault can't be used for it as it is withdrawn as ETH.

only WETH from an existing uniswap position could be used for this.  One potential solution is wrapping ETH balance before the callback and unwrapping weth after transfering the amount owed from the flashswap.

aleone: Another potential solution is to send some ETH to uniswap as part of the call or repay with ETH, this possible?

aleone (changes requested):
See comments on wrapping weth, swaps are limited from ETH to oSQTH as we don't wrap any withdrawn ETH or payable ETH to weth.

packages/hardhat/contracts/periphery/ControllerHelper.sol
haythem96: I will just wrap the contract balance again, also to be consistent by repaying WETH to pool (not sending ETH)

aleone (approved):
lgtm

PR #331, opened 2022-04-11
Add unsupported network

### LINKED ISSUE #162: Don't let users use the site on unsupported chains
User tried to use the site from Avalanche and you're able to submit txns from other EVM chains even though they're invalid
Add a blocker to prevent users from doing that

block902:
# Task: Add unsupported network and disable all trades on unsupported chains.

## Description
- Add unsupported network button on header
- Disable all trades on unsupported chains.

Fixes #162 

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/packages/frontend/pages/positions.tsx
+++ b/packages/frontend/pages/positions.tsx
@@ -151,15 +151,20 @@ const PositionsHome = () => {
 
 const ConnectWallet: React.FC = () => {
   const selectWallet = useSelectWallet()
+  const supportedNetwork = useAtomValue(supportedNetworkAtom)
   const classes = useStyles()
 
   return (
     <div>
       <Nav />
       <div className={classes.container}>
-        <LinkButton style={{ margin: 'auto' }} onClick={selectWallet}>
-          Connect Wallet
-        </LinkButton>
+        {supportedNetwork ? (
+          <LinkButton style={{ margin: 'auto' }} onClick={selectWallet}>
+            Connect Wallet
+          </LinkButton>
+        ) : (
+          <Typography>Unsupported Network</Typography>
+        )}
       </div>
     </div>
   )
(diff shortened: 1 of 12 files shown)


EstherSeyi (approved):

yichiehliu (commented):
Do we also add !supportedNetwork on tx buttons in vault page to disable vault actions(mint, burn, lp nft, add collat, remove collat)? Otherwise I can still type in input and click on submit tx
image

KMKoushik (changes requested):
Good catch @yichiehliu 

@block902 Well this fix kinda works, it can still miss some places now or in the future. Is there any other better way to do this? 

Like even if we call the function it should cancel the call or something like that.

packages/frontend/pages/vault/[vid].tsx
```diff
@@ -930,34 +937,40 @@ const Component: React.FC = () => {
...
+                          <CircularProgress color="primary" size="1rem" />
+                        ) : (
+                          'Mint'
+                        )}
+                      </AddButton>
+                    </div>
+                  )}
                 </div>
               </div>
               <div className={classes.manager}>
```
Comment (yichiehliu) [inline packages/frontend/pages/vault/[vid].tsx:976]: I think we also need to disable lp nft select when user is on unsupported network

yichiehliu (changes requested):

packages/frontend/pages/vault/[vid].tsx
Comment (block902) [reply packages/frontend/pages/vault/[vid].tsx:981]: Fixed this and pushed a commit

KMKoushik:
> > Good catch @yichiehliu
> > @block902 Well this fix kinda works, it can still miss some places now or in the future. Is there any other better way to do this?
> > Like even if we call the function it should cancel the call or something like that.
> 
> I think it would be better to disable the buttons to make transactions on unsupported networks

Disabling is fine. But what if we forget to do it in some new places. The ultimate goal of the fix should avoid user loosing money IMO

packages/frontend/pages/positions.tsx
```diff
@@ -225,355 +227,393 @@ export function Positions() {
           <Typography color="primary" variant="h6">
             Your Positions
           </Typography>
-          <div style={{ display: 'flex' }}>
-            <Typography component="span" color="textSecondary">
-              ETH Price:{' '}
-            </Typography>
+          {supportedNetwork && (
```
KMKoushik, line 230: Same here, What if we forget this in some places? Instead of manually adding this check in every possible component we should just reset the address, so there will be no data. Remembering this is will definitely be difficult.

@yichiehliu @EstherSeyi @adam5909099 Thoughts

KMKoushik (changes requested):

KMKoushik:
@block902 Seems the changes requested are not being done. What's the plan for this? Do you have any doubts about implementation?

KMKoushik:
@block902 Let me know once you added the requested changes

KMKoushik (approved):

Issue #300, opened 2022-04-06
LP rebalances that either trade or mint may not get the amount of oSQTH correct, can we fix?


aleone:
Is there a way to ensure that the amount of oSQTH purchased (or sold) is the exact amount required, maybe by checking w uniswap nft manager some way?  Similar to [#299](https://github.com/opynfinance/squeeth-monorepo/issues/299), but making a new issue for the rebalance case.   Ideally we should be able to know the exact amount needed before we mint/buy/etc instead of using the f/e provided numbers and needing to burn/send back excess.

Issues basically anywhere we call checkExcess

haythem96:
fixed

PR #360, opened 2022-04-20
Fix excess amounts when LPing

### LINKED ISSUE #299: LPing can leave excess oSQTH in the contract (which is burned or returned), can we prevent this from happening?
When you LP on uniswap you target an amount of oSQTH and ETH to add, but its possible you need less of one or the other based on the current market price.

Current implementation:
1. check if there is excess oSQTH and burn it or return to user if no vault
2. return ETH / WETH to user 

Better fix:
1. mint the exact amount of oSQTH that is needed to LP (is this what is done in the swap and add functionality that uniswap has?)

Is this something that we can fix, right now it will make a lot of transactions cost significantly more, as we will often not get the exact asset composition correct on the front end.

This is a problem in:
- batchMintLp

haythem96:
# Task: Fix excess amounts when LPing

## Description

Fixes #299 #300 #401 #402

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/packages/hardhat/contracts/periphery/ControllerHelper.sol
+++ b/packages/hardhat/contracts/periphery/ControllerHelper.sol
@@ -468,9 +468,6 @@ contract ControllerHelper is UniswapControllerHelper, EulerControllerHelper, IER
         uint8 _callSource,
         bytes memory _calldata
     ) internal override {
-        // convert flashloaned WETH to ETH
-        // IWETH9(ControllerHelperDiamondStorage.getAddressAtSlot(5)).withdraw(_amount);
-
         if (
             ControllerHelperDataType.CALLBACK_SOURCE(_callSource) ==
             ControllerHelperDataType.CALLBACK_SOURCE.FLASHLOAN_W_MINT_DEPOSIT_NFT
--- a/packages/hardhat/contracts/periphery/ControllerHelper.sol
+++ b/packages/hardhat/contracts/periphery/ControllerHelper.sol
@@ -57,6 +57,8 @@ contract ControllerHelper is UniswapControllerHelper, EulerControllerHelper, IER
 
         IWPowerPerp(IController(_controller).wPowerPerp()).approve(_nonfungiblePositionManager, type(uint256).max);
         IWETH9(IController(_controller).weth()).approve(_nonfungiblePositionManager, type(uint256).max);
+
+        INonfungiblePositionManager(_nonfungiblePositionManager).setApprovalForAll(_controller, true);
     }
 
     /**
@@ -597,6 +599,11 @@ contract ControllerHelper is UniswapControllerHelper, EulerControllerHelper, IER
                         increaseLiquidityParam,
                         isWethToken0
                     );
+
+                    IController(ControllerHelperDiamondStorage.getAddressAtSlot(0)).depositUniPositionToken(
+                        vaultId,
+                        increaseLiquidityParam.tokenId
+                    );
                 } else if (
                     data[i].rebalanceVaultNftType == ControllerHelperDataType.RebalanceVaultNftType.DecreaseLpLiquidity
                 ) {
@@ -677,10 +684,6 @@ contract ControllerHelper is UniswapControllerHelper, EulerControllerHelper, IER
                     );
 
                     // deposit Uni NFT token in vault
-                    INonfungiblePositionManager(ControllerHelperDiamondStorage.getAddressAtSlot(6)).approve(
-                        ControllerHelperDiamondStorage.getAddressAtSlot(0),
-                        tokenId
-                    );
                     IController(ControllerHelperDiamondStorage.getAddressAtSlot(0)).depositUniPositionToken(
                         vaultId,
                         tokenId
--- a/packages/hardhat/contracts/periphery/ControllerHelper.sol
+++ b/packages/hardhat/contracts/periphery/ControllerHelper.sol
@@ -9,7 +9,6 @@ import "hardhat/console.sol";
 import {IWETH9} from "../interfaces/IWETH9.sol";
 import {IWPowerPerp} from "../interfaces/IWPowerPerp.sol";
 import {IShortPowerPerp} from "../interfaces/IShortPowerPerp.sol";
-import {IOracle} from "../interfaces/IOracle.sol";
 import {IController} from "../interfaces/IController.sol";
 import {IUniswapV3Pool} from "@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol";
 import {IERC721Receiver} from "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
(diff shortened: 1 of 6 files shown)


packages/hardhat/contracts/periphery/lib/ControllerHelperUtil.sol
```diff
@@ -134,31 +168,38 @@ library ControllerHelperUtil {
...
-            __wPowerPerpToMint,
+            _wPowerPerpToMint,
             0
         ));
     }
 
     /**
      * @notice burn wPowerPerp or just withdraw collateral from vault (or both)
      * @param _controller controller address
+     * @param _weth WETH address
```
alpinechicken, line 189: x2

aleone:
Can we rebase the changes from the wrap/unwrap branch, this is showing those changes as well. Or maybe merge that PR and rebase from main.

packages/hardhat/contracts/periphery/lib/ControllerHelperUtil.sol
```diff
@@ -71,14 +79,41 @@ library ControllerHelperUtil {
...
      * @return _vaultId and tokenId
      */
-    function mintAndLp(address _controller, address _nonfungiblePositionManager, address _wPowerPerp, address _wPowerPerpPool, ControllerHelperDataType.MintAndLpParams calldata _mintAndLpParams, bool _isWethToken0) public returns (uint256, uint256) {
+    function mintAndLp(address _controller, address _nonfungiblePositionManager, address _wPowerPerp, address _wPowerPerpPool, address _weth, address _oracle, ControllerHelperDataType.MintAndLpParams calldata _mintAndLpParams, bool _isWethToken0) public returns (uint256, uint256) {
+        IWETH9(_weth).withdraw(_mintAndLpParams.collateralToDeposit);
+
+        uint256 amount0Desired; 
+        uint256 amount1Desired;
+        {
+            int24 timeWeightedAverageTick = IOracle(_oracle).getTimeWeightedAverageTickSafe(_wPowerPerpPool, TWAP_PERIOD);
```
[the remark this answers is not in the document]
aleone: https://github.com/Uniswap/v3-periphery/blob/main/contracts/base/LiquidityManagement.sol#L67

packages/hardhat/contracts/periphery/lib/ControllerHelperUtil.sol
```diff
@@ -71,14 +79,41 @@ library ControllerHelperUtil {
...
+            uint160 sqrtRatioX96 = TickMathExternal.getSqrtRatioAtTick(timeWeightedAverageTick);
+            uint160 sqrtRatioAX96 = TickMathExternal.getSqrtRatioAtTick(_mintAndLpParams.lowerTick);
+            uint160 sqrtRatioBX96 = TickMathExternal.getSqrtRatioAtTick(_mintAndLpParams.upperTick);
+            (amount0Desired, amount1Desired) = _isWethToken0 ? (_mintAndLpParams.collateralToLp, _mintAndLpParams.wPowerPerpAmount) : (_mintAndLpParams.wPowerPerpAmount, _mintAndLpParams.collateralToLp);
+            uint128 maxLiquidity = LiquidityAmounts.getLiquidityForAmounts(uint160(timeWeightedAverageTick), sqrtRatioAX96, sqrtRatioBX96, amount0Desired, amount1Desired);
```
aleone, line 93: Some of tests are failing because ```timeWeightedAverageTick``` is used as a param instead of ```sqrtRatioX96``` here.  

The other tests are failing as the ticks are not correct (I'm pretty sure).  We can just use 0 as the dividing tick for all oSQTH or all ETH, a tick of 0 corresponds to a price of 1.0001^0=1, where a tick of 1000 corresponds to a price of 1.0001^1000 = 1.1, etc.

I revamped the process in the integration test file to calculate all of this correctly as it was a problem there as well.

aleone (changes requested):
Think I've figured out why each of tests fail (2 main reasons) and have noted them in comments.

packages/hardhat/contracts/periphery/lib/ControllerHelperUtil.sol
aleone: Code for how I fixed the tests for LP of specified composition: https://github.com/opynfinance/squeeth-monorepo/blob/main/packages/hardhat/test/integration-tests/periphery/controller-helper.ts#L1471

packages/hardhat/contracts/periphery/lib/ControllerHelperUtil.sol
```diff
@@ -10,16 +12,22 @@ import {IController} from "../../interfaces/IController.sol";
...
+import {IOracle} from "../../interfaces/IOracle.sol";
 
 // lib
 import {SafeMath} from "@openzeppelin/contracts/math/SafeMath.sol";
-import {ControllerHelperDataType} from "./ControllerHelperDataType.sol";
 import {Address} from "@openzeppelin/contracts/utils/Address.sol";
+import {ControllerHelperDataType} from "./ControllerHelperDataType.sol";
+import {LiquidityAmounts} from "./LiquidityAmounts.sol";
+import {TickMathExternal} from "../../libs/TickMathExternal.sol";
+import {SqrtPriceMathPartial} from "../../libs/SqrtPriceMathPartial.sol";
```
alpinechicken, line 23: do we use this?

alpinechicken (approved):
lgtm

packages/hardhat/contracts/periphery/lib/ControllerHelperUtil.sol
haythem96: Cleaned it

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -628,19 +636,19 @@ contract ControllerHelper is UniswapControllerHelper, EulerControllerHelper, IER
                     );
                 } else if (
                     // this will execute if the use case is to mint in vault, deposit collateral in vault or mint + deposit
-                    data[i].rebalanceVaultNftType == ControllerHelperDataType.RebalanceVaultNftType.MintIntoVault
+                    data[i].rebalanceVaultNftType == ControllerHelperDataType.RebalanceVaultNftType.DepositIntoVault
```
aleone, line 639: Dont love the name (new or old), maybe DepositMintVault or something, but not critical.

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -660,25 +668,22 @@ contract ControllerHelper is UniswapControllerHelper, EulerControllerHelper, IER
...
-                    ControllerHelperDataType.LpWPowerPerpPool memory mintNewLpParams = abi.decode(
+                    ControllerHelperDataType.MintAndLpParams memory mintAndLpParams = abi.decode(
                         data[i].data,
-                        (ControllerHelperDataType.LpWPowerPerpPool)
+                        (ControllerHelperDataType.MintAndLpParams)
                     );
 
-                    uint256 tokenId = ControllerHelperUtil.lpWPowerPerpPool(
+                    uint256 tokenId;
+                    (vaultId, tokenId) = ControllerHelperUtil.mintAndLp(
```
aleone, line 677: I don't think this works because here you may have a balance of oSQTH and ETH from removing liquidity from uniswap.  Think the mintAndLp will mint the total needed to LP assuming 0 internal balance of oSQTH.  I suggested some ideas around more global variables that can be set by certain actions, could be a solution for this..lmk if makes sense.

aleone (approved):
Generally looks good. A few things I caught, that need to be fixed, either here or in other PR, up to you.

- IncreaseLpLiquidity still relies on checkExcess, it should have the same logic as mint and LP functions (#402)
- MintNewLp doesn't respect the oSQTH balance of the contract so will always mint too much oSQTH if there was any withdrawn from LPing (#401)

aleone (approved):
Generally looks good, but I think we should replace the diamond storage that points to wPowerPerpPool to a param so people can LP (and close, and rebalance, etc) non 0.3% pools.  I think its pretty straightforward to support this, in the same way that you have allowed swaps to different pools and truly makes the ControllerHelper agnostic to the uniswap pool chosen.

Issue #220, opened 2022-03-08
Long/Close didn't load a user's oSQTH balance


alexisgauba:
**Describe the bug**
User went to long/close on the trade page to close their long squeeth position, but it didn't register their balance. They were concerned until they went to Uniswap, saw their balance, and closed out their position from there

**To Reproduce**

1. Open a long position buy buying squeeth from long/open 
2. Leave the page 
3. Go back to squeeth.com and try to close oSQTH position by selling from long/close on the trade page 

**Expected behavior**
Should load user's oSQTH balance letting them close from long/close

yichiehliu:
It still happens to me sometimes, so reopened the issue
https://user-images.githubusercontent.com/26922518/165255039-6f77976a-59d1-4413-a185-7ec1a4308b14.mov

PR #385, opened 2022-04-26
Long/Close didn't load a user's oSQTH balance

EstherSeyi:
# Task:

## Description

- Adds activity update for when sqthBal is loading or when there is an error loading the sqth bal.
- Allow user to refetch incase sqth bal fails to load. 

Fixes # (issue)
#220 

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/packages/frontend/src/components/Trade/Long/index.tsx
+++ b/packages/frontend/src/components/Trade/Long/index.tsx
@@ -247,7 +247,7 @@ const useStyles = makeStyles((theme) =>
   }),
 )
 
-const OpenLong: React.FC<BuyProps> = ({ activeStep = 0, open, sqthBalProps }) => {
+const OpenLong: React.FC<BuyProps & LongProps> = ({ activeStep = 0, open, sqthBalProps }) => {
   const [buyLoading, setBuyLoading] = useState(false)
   const getBuyQuoteForETH = useGetBuyQuoteForETH()
   const getBuyQuote = useGetBuyQuote()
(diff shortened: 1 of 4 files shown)


yichiehliu (commented):
is it possible to make long squeeth balance an atom? cuz the oSQTH balance in trade card is still not updated after a close trade on long side, still required a refresh or click on other tabs to get the correct balance

yichiehliu (changes requested):
Were you able to test with long close? cuz the tx would fail for me

Uploading 螢幕錄製 2022-04-29 上午5.05.23.mov…

yichiehliu:
ok, the tx succeeded after several hard refreshes, not sure if it's related to the changes

packages/frontend/src/components/Trade/Long/index.tsx
```diff
@@ -960,11 +985,31 @@ type BuyProps = {
...
 const Long: React.FC<BuyProps> = ({ open, isLPage = false, activeStep = 0 }) => {
+  const { longSqthBal, refetch: refetchSqthBal, loading: loadingSqthBal, error } = useLongSqthBal()
+
   return open ? (
-    <OpenLong open={open} isLPage={isLPage} activeStep={activeStep} />
+    <OpenLong
+      open={open}
+      isLPage={isLPage}
+      activeStep={activeStep}
+      sqthBalProps={{ longSqthBal, refetch: refetchSqthBal, loading: loadingSqthBal }}
```
yichiehliu, line 1005: I noticed there are some inconsistencies for long balance in the codebase, cuz there is also long balance in the osqth input of <OpenLong /> component, but the balance will change right after the open long trade finished, it gets updated with the same value on position card and at the same time. 

We dont have balance not loaded issue on buy side, cuz we r using [squeethAmount variable ](https://github.com/opynfinance/squeeth-monorepo/pull/385/files#diff-0df6f866c3e8aa6fd371a3fdb870c356e099f912847729a6c7693c57597901fcR511) as our long balance there, same as the value on position card and position page, but on long close, now we r using useLongSqthBal, which does some extra fetching and this hook is only used in long trade card, but think squeethAmount already did the fetching job to get correct position value. So we dont need extra fetching in long close component

So thinking it's better to have a jotai atom for long position value across the app, or at least we all use same value / variable for same position

yichiehliu (changes requested):

yichiehliu (approved):

adam5909099 (approved):

KMKoushik (approved):

PR #388, opened 2022-04-27
Fix: vault auth check

### LINKED ISSUE #387: Access control not implemented for all functions
Access control is not implemented for all functions. We need to check to make sure people are acting on vaults they own if not vault 0 (vault 0 is valid for those functions that can create a new vault).

haythem96:
# Fix: vault auth check

## Description

Fixes #387 

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist


aleone:
I think this function is fine, not sure where I saw the issue now.  I think the original implementation is fine, isnt it?

We don't ever need id 0 for rebalanceVaultNft right, so can go back to original (pre this commit)?

haythem96:
At a first look I wasn't sure where the issue is, as all function have the correct auth check.
But for `rebalanceVaultNft()` isn't vaultId = 0 needed if user want to use `RebalanceVaultNftType.MintNewLp` action in a new vault ?

packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -574,6 +571,13 @@ contract ControllerHelper is UniswapControllerHelper, EulerControllerHelper, IER
                 (uint256, ControllerHelperDataType.RebalanceVaultNftParams[])
             );
 
+            // check ownership
```
alpinechicken, line 574: nice I think this fixes my problem too

alpinechicken (approved):
lgtm

alpinechicken (approved):
lgtm

aleone:
> At a first look I wasn't sure where the issue is, as all function have the correct auth check. But for `rebalanceVaultNft()` isn't vaultId = 0 needed if user want to use `RebalanceVaultNftType.MintNewLp` action in a new vault ?

I don't really see that people will/should use this function for new vaults (we have others), but don't see a big issue including it to let people use it for new vaults.  

However, I think we should do the check in the external function vs in the callback, like it originally was, and you can add the if vaultId>0 check to what was there before. Any reason you moved it?

haythem96:
Not needed

PR #400, opened 2022-04-29
Add Uniswap pool fee as functions arg

### LINKED ISSUE #294: Update ControllerHelper pools to take a pool input from f/e to allow multi pool routing in future (pool, fee, etc)
If there is ever liquidity in different fee tier pools, this contract can not support trading other than the 0.3% pool, maybe should be included, or could be added / changed in future.

haythem96:
# Add Uniswap pool fee as functions arg

## Description

This will allow choosing different uniswap pool from fe/user.

Fixes #294 

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ workflows:
             - compile
       - coverage:
           requires:
-            - compile
+            - e2e-test
           context: BUILD_KEYS
       - contract-size:
           requires:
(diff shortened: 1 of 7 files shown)


aleone (approved):
lgtm generally, but think we should add in logic to replace the diamond storage for wPowerPerpPool and have it be a param, like the pool fee for at least non vault based actions (batchMintLp, rebalanceWithoutVault, etc).  That way someone can use the ControllerHelper to LP in non 0.3% pools.  Seems pretty straightforward and makes the ControllerHelper fully agnostic to pools (LPing and trading). Note that (if easier) it doesn't need to support arbitrary pools for vault based actions (flashloan mint and LP for example) as they will fail as Controller will not accept them as collateral.

PR #404, opened 2022-04-29
Fix burn bug

### LINKED ISSUE #383: Users that want to burn the exact amount of oSQTH withdrawn from a vault may have transactions fail
One use case of ControllerHelper is that users can withdraw some liquidity from their LP position and burn an amount of oSQTH against their vault.  When a user specifies an amount of oSQTH to burn and the LP amount returned is different, a trade occurs to either sell the excess of buy any deficit. A user specifies a limit price for this.  This is good when a user wants to trade, but if a user wants to use the exact amount of oSQTH withdrawn from the LP to burn, they will often get a small residual amount of difference that will need to be traded (due to rounding or change in pool price from time of sending tx to mining).  

If the amount if 50% of the time higher and 50% of the time lower, 50% of the time a users transaction will fail as the slippage is the wrong direction.  To be clear, this only impacts users that attempt to exactly specify an amount to burn that is equal to the amount of oSQTH they will receive from the LP.  The problem is that the front end can never know with certainity what this will be.

@alexisgauba @zkoticha - I think this is fairly common edge case/use case that we may want, so should probably fix this.  Lmk thoughts.

Some solutions:

Hacky solution:
- change the slippage to be in term of 2 params:
  - ```spotPriceEthPowerPerp``` is the f/e spot price
  - slippage is a % that the user has specified that are willing to pay above ```spotPriceEthPowerPerp``` if buying or sell below spotPriceEthPerpPowerPerp if selling
  - the use case of wanting to exactly burn the amount withdrawn isn't fixed in this case, users may often do small trades with the pool

One less hacky, better solution:
- change the function to have a ```burnExactRemoved``` flag
  - this flag skips the trading logic and burns the exact amount removed from the LP

haythem96:
# Fix burn bug

## Description

Fixes #383

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/packages/hardhat/contracts/periphery/ControllerHelper.sol
+++ b/packages/hardhat/contracts/periphery/ControllerHelper.sol
@@ -473,9 +473,6 @@ contract ControllerHelper is UniswapControllerHelper, EulerControllerHelper, IER
         uint8 _callSource,
         bytes memory _calldata
     ) internal override {
-        // convert flashloaned WETH to ETH
-        // IWETH9(ControllerHelperDiamondStorage.getAddressAtSlot(5)).withdraw(_amount);
-
         if (
             ControllerHelperDataType.CALLBACK_SOURCE(_callSource) ==
             ControllerHelperDataType.CALLBACK_SOURCE.FLASHLOAN_W_MINT_DEPOSIT_NFT
--- a/packages/hardhat/contracts/periphery/ControllerHelper.sol
+++ b/packages/hardhat/contracts/periphery/ControllerHelper.sol
@@ -199,7 +199,6 @@ contract ControllerHelper is UniswapControllerHelper, EulerControllerHelper, IER
             _params.collateralToWithdraw,
             _params.limitPriceEthPerPowerPerp
         );
-        wrapInternal(address(this).balance);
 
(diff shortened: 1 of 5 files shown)


packages/hardhat/contracts/periphery/ControllerHelper.sol
```diff
@@ -888,40 +890,52 @@ contract ControllerHelper is UniswapControllerHelper, EulerControllerHelper, IER
         uint256 _wPowerPerpAmount,
```
alpinechicken, line 890: missing natspec

alpinechicken (approved):
lgtm - I think a bit confusing because wPowerPerpToBurn is unused if burnExactRemoved is true but can't be helped

aleone (approved):
lgtm, but don't see a way for a user to do this type of action in rebalance with vault actions (burn exact amount of oSQTH withdrawn from LP). Think this should be added as well.

haythem96:
> lgtm, but don't see a way for a user to do this type of action in rebalance with vault actions (burn exact amount of oSQTH withdrawn from LP). Think this should be added as well.

But there will be no trade in those one right ?

Issue #370, opened 2022-04-25
When some users partial close it is to target a certain amount of ETH in their wallet which requires guess + check in current flow


alexisgauba:
**Please describe the problem you're having.**
For some users, when they are partially closing their vault it is to target having a specific amount of ETH collateral left in the vault after closing vs. a specific amount of oSQTH to burn. Right now they need to guess & check toggle the oSQTH to buy back + burn and collatRatio to get to a desired amount of ETH vs. being able to select the desired ETH amount

PR #428, opened 2022-05-05
Add collateral you redeem

adam5909099:
# Task:
This PR involves refactoring the existing logic of ShortClose so needs a detailed review.

## Description
Add collateral Eth input and refactored dependency logic 
image

Fixes https://github.com/opynfinance/squeeth-monorepo/issues/370

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist


KMKoushik:
@adam5909099 Can you explain what has been done clearly? PRs like this might have to go through @alexisgauba and user story should be documented!

adam5909099:
@alexisgauba, Can you please review this PR for allowing users to enter collateral? I kept the Spend card for now because I'm not sure how to label the two values contained in the Card.

alexisgauba:
Yep, Adam and I chatted about this earlier! Main thing is giving users the ability to input the amount of ETH they want to remove from a vault when they are partially closing, some advanced users would just use amount to buy back + burn & collRatio to guess and check amount of ETH to remove, when what they wanted to input was a certain amount of ETH to get out and then see how that affects their position 

Adding the field for Collateral to redeem the way you did LGTM visually 

One good thing to add would be the liquidation price of the vault after the change. The key things advanced users were looking at were 
- How much ETH can I remove from the vault 
- What liquidation price will that leave me at  

Imo could just add "New liquidation price" as a row right above "Current collateral ratio" 

From a logic pov, seems like the value is a bit different than on the current deployment? Seems like something might be going on there? 
Screen Shot 2022-05-06 at 2 59 00 PM
Screen Shot 2022-05-06 at 2 59 12 PM

KMKoushik:
@adam5909099 Can you take a look at this?

adam5909099:
@alexisgauba For different collateral value between this PR and production isn't an issue. It's related to the time when the website is loaded.

1.
image
2.
image

The above screenshots are from production when closing two oSQTH in short. They show different values because they were loaded at other times. So I don't think it's an issue.

CC @KMKoushik

alexisgauba:
Ok gotcha, then in that case I think the main think would just be adding "New liquidation price" as a row right above "Current collateral ratio"

adam5909099:
@alexisgauba, I got it, thank you. I added a new liquidation price, but I don't think it's calculated correctly. Can you or @KMKoushik provide me the formula for calculating it in a short close or a notion doc that I can reference?

adam5909099:
@KMKoushik I updated PR as you mentioned. When I set 0.2869 collateral eth, the new liquidation price was 4866.64, like in the following screenshot.
image
After partially closing short (it took about 15 seconds) and going to the manage vault page, the liquidation price was shown as 4866.63
image
There's a 0.01 difference, but I think it's related to underlying value changes while closing short and not an issue.

KMKoushik (approved):

KMKoushik (changes requested):
Can we stop having -ve value here? We can stop slider when it reaches 0
Screen Shot 2022-05-13 at 1 44 57 pm

adam5909099:
@KMKoushik I add a restriction to the slider to ensure eth is bigger than 0. Can you please review it again?

PR #476, opened 2022-05-18
Include flashloan amount in collateralToDeposit to allow user to use vault collateral to LP

### LINKED ISSUE #399: For an existing vault w flashloanWMintDepositNft, we withdraw collat at end so a user cant use vault collateral for LPing
I think withdrawing the collateral at the start of the transaction is a better pattern as a user may want to use their collateral for LPing.  A use case would be someone has a 300 CR short with a bunch of ETH collateral and want to mint + LP against the vault using some of the excess vault collateral for it. Most of vault collateral can be used if the LP token is deposited as collateral.  It may be impossible to use the ETH as collateral unless we change the pattern of withdrawing.

haythem96:
# Fix: Include flashloan amount in collateralToDeposit to allow user to use vault collateral to LP

## Description

This PR remove adding the amount to flashloan to the collateral to deposit when minting and LP, there this will allow the user to user any extra amount flashloaned as amount to LP,  and will repaid later when withdrawing from vault.

In the function call, the amount to flashloan and to use as vault collateral should be added in the collateral to deposit param.

Fixes #399 

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist

### DIFF
--- a/packages/hardhat/contracts/periphery/ControllerHelper.sol
+++ b/packages/hardhat/contracts/periphery/ControllerHelper.sol
@@ -490,7 +490,7 @@ contract ControllerHelper is UniswapControllerHelper, EulerControllerHelper, IER
                     wPowerPerpPool: ControllerHelperDiamondStorage.getAddressAtSlot(3),
                     vaultId: data.vaultId,
                     wPowerPerpAmount: data.wPowerPerpAmount,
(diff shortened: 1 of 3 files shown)


alpinechicken (approved):
lgtm

aleone (approved):
lgtm

PR #477, opened 2022-05-18
Fix: general swap fail when amountIn is greater than balance

### LINKED ISSUE #435: General swaps will often fail if trying to do exact oSQTH or ETH withdrawn swaps
This type of trade will likely fail often for a live trade as we don't know the exact amount we will need to sell.  If f/e thinks we will get out 100 oSQTH and tries to swap it all, but we only get 99.99, this will fail.  Same for ETH and selling to oSQTH.

One solution is that we use the min(balanceOf(address.this), wPowerPerpToSell param)

haythem96:
# Fix: general swap fail when amountIn is greater than balance

## Description

Fixes #435 

## Type of change

- [ ] Document update

## How Has This Been Tested

## Checklist


aleone (approved):
lgtm

alpinechicken (approved):
lg

PR #486, opened 2022-05-19
Refactor computed swaps

adam5909099:
# Task:
Investigate sharing effects between components without creating context manually.

## Description
This PR refactors computed swaps to use constate rather than handling context API manually. Using this approach, we can easily update all existing hooks that need effects sharing to avoid duplicated effects call. (It'll improve performance by reducing API calls and duplicated calculation)


KMKoushik:
@EstherSeyi @yichiehliu Can you check this PR?

packages/frontend/src/state/positions/hooks.ts
```diff
@@ -112,15 +113,115 @@ export const useSwaps = () => {
...
+      setPositionType(PositionType.NONE)
+    }
+  }, [computedSwaps.squeethAmount, oSqueethBal, setPositionType, setIsToHidePnL])
 
-  return context
-}
+  useAppEffect(() => {
+    refetch()
+  }, [computedSwaps.squeethAmount, refetch])
+
```
EstherSeyi, line 210: @adam5909099  there is this continuous re-rendering that's happening. I think it's coming from these lines 207-209. Cos when I comment those lines, it stops.
https://user-images.githubusercontent.com/40301524/169861910-1179b90d-63b6-4fa4-9d4f-3cdd40a8b779.mp4

KMKoushik: @adam5909099 Seems to be an issue, can you check this!

adam5909099: @EstherSeyi I tried logging in that useEffect callback but it's not called infinitely
image
image
Can you let me know the step for reproducing your recording?

yichiehliu: It seems that when you have longer txs history, it will cause this loop 
My squeeth amount was not changed here with this address `0x49A8E210967553fd22a0C235d47cFAef321276f5`

https://user-images.githubusercontent.com/26922518/170023318-c0f8c8d8-f686-4239-964b-d629a917e6a3.mov

EstherSeyi: @adam5909099 I wanted to add to my short position, so I was on the short open tab when it was happening.

adam5909099: @yichiehliu Please pull the latest of this branch. I can see the console log from providers.tsx, but the providers component doesn't exist anymore

adam5909099: @EstherSeyi It was related to polling in useTokenBalance. I resolved this by adding queryRefecth, which handles real refetching.

EstherSeyi (approved):

packages/frontend/src/state/positions/hooks.ts
```diff
@@ -112,15 +113,117 @@ export const useSwaps = () => {
...
+      setPositionType(PositionType.SHORT)
+    } else {
+      setIsToHidePnL(false)
+      setPositionType(PositionType.NONE)
+    }
+  }, [computedSwaps.squeethAmount, oSqueethBal, setPositionType, setIsToHidePnL])
 
-  return context
-}
+  console.log('refetching ....')
```
yichiehliu, line 207: Do we remove this before merging?

PR #490, opened 2022-05-22
Chore: Make variables clearer in rebalanceWithoutVault for LP amounts

aleone:
refactors the return from getAmountsToLp into an if else statement so variables don't have to be swapped

8 gas difference


haythem96 (approved):

alpinechicken (approved):
lgtm

PR #517, opened 2022-06-14
Init crab v2

KMKoushik:
Task: 

Forked Crab contract to Crabv2 contract. This will be the base contract to work on

### DIFF
--- a/packages/hardhat/contracts/strategy/CrabStrategyV2.sol
+++ b/packages/hardhat/contracts/strategy/CrabStrategyV2.sol
@@ -1,0 +1,1126 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+pragma solidity =0.7.6;
+pragma abicoder v2;
+
+// interface
+import {IController} from "../interfaces/IController.sol";
+import {IWPowerPerp} from "../interfaces/IWPowerPerp.sol";
+import {IOracle} from "../interfaces/IOracle.sol";
+import {IWETH9} from "../interfaces/IWETH9.sol";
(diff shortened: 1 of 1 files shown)


aparnakr:
why is the e2e test failing? Is that an issue?

sanandnarayan:
@aparnakr @KMKoushik If we are starting a new contract. Shall we use the latest version of solidity?

KMKoushik:
> why is the e2e test failing? Is that an issue?
> 
> Could you also port tests over to work with the v2 strat?

@aparnakr That's an existing thing. It fails in CI @haythem96 Is working on it

KMKoushik:
@aparnakr @sanandnarayan Ported V1 tests to V2. @sanandnarayan Not changing the version of sol as uniswap contracts doesn't work in newer version.

aleone (changes requested):
Don't we need to update all of these copied tests to point to CrabStrategyV2. Right now aren't they just testing the CrabStrategy (ie v1) contract.  IMO we add functionality and edit/remove/add stuff from the v2 tests, but they will always pass if pointing to old contract.

KMKoushik:
> Don't we have to update the typechain import to point to CrabStrategyV2 in every test? Right now (I think) it works because the functions are all the same, but in future it wont.
> 
> ```
> import { WETH9, MockErc20, Controller, Oracle, WPowerPerp, CrabStrategy, ISwapRouter } from "../../../typechain";
> ```

@aleone Done

aleone (approved):

aparnakr:
can we merge this?

PR #518, opened 2022-06-14
Add transfer vault capability for crab v2

KMKoushik:
# Task:
Add vault transferring capability to crabV2

I am getting `Contract code size exceeds 24576 bytes` so I disabled it for now and we have to fix it later

## Description

- [x] Add `transferVault` event
- [X] Set Cap to 0
- [X] Add tests for transfer vault
- [X] Add time lock contract
- [x] Add tests for timelock


aparnakr (approved):
Can you also add a simple test for this?

aparnakr:
Do we want to add a timelock as well? If so you could look at https://github.com/compound-finance/compound-protocol/blob/master/contracts/Timelock.sol

KMKoushik:
> Do we want to add a timelock as well? If so you could look at https://github.com/compound-finance/compound-protocol/blob/master/contracts/Timelock.sol

@aparnakr How bout making a separate PR for this?

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -193,15 +195,29 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
+        emit VaultTransferred(_newStrategy, vaultId);
+    }
+
     /**
      * @notice owner can set the strategy cap in ETH collateral terms
      * @dev deposits are rejected if it would put the strategy above the cap amount
      * @dev strategy collateral can be above the cap amount due to hedging activities
      * @param _capAmount the maximum strategy collateral in ETH, checked on deposits
      */
     function setStrategyCap(uint256 _capAmount) external onlyOwner {
```
aleone, line 215: Might be more simple to make this be a public function instead of having an external that calls internal function?  Get why you changed it, but public can have same result, probably with smaller bytecode and small gas differences (it may be slightly more expensive to call?)

I don't have strong opinions though, other thoughts from folks?  Its also possible we just keep in mind for future if we have bytecode size issues later on and can change it.

packages/hardhat/hardhat.config.ts
```diff
@@ -74,6 +74,8 @@ const config: HardhatUserConfig = {
       accounts: {
         accountsBalance: '1000000000000000000000000000'
       },
+      // IMPORTANT: SHOULD BE REMOVED AND FIXED BEFORE MERGING
```
aleone, line 77: Easier way to fix this temp issue is to set the ```runs: 800``` under optimizer to 1 and see if that gets you under the limit and can raise it when we remove code.

aleone (approved):
Generally lgtm, few small comments. Ok doing the timelock in separate PR, but do think it may change things a decent amount / larger scope. Fine with either way of doing it.  

TLDR:
- Think easier to use optimizer to fix contract bytecode size for now, reverted later
- add missing natspec
- could consider public vs external/internal pattern especially if hit bytecode limit later

packages/hardhat/hardhat.config.ts
aleone: (spoiler, it does get us under the limit :) )

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
KMKoushik: The only reason I used `external` is that I read it's cheap. But yeah it's not frequently called function and doesn't necessarily have to be external. Will change!

packages/hardhat/hardhat.config.ts
KMKoushik: Yep worked

KMKoushik:
@aparnakr @aleone Lot's of changes needed in vault transfer logic to implement timelock. So going to use the same PR. Not ready yet

aparnakr:
> > Do we want to add a timelock as well? If so you could look at https://github.com/compound-finance/compound-protocol/blob/master/contracts/Timelock.sol
> 
> @aparnakr How bout making a separate PR for this?

yes makes sense

packages/hardhat/contracts/strategy/timelock/SafeMath.sol
```diff
@@ -0,0 +1,191 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
KMKoushik, line 1: Don't want to change any code from the compound, so I copied math lib too

packages/hardhat/hardhat.config.ts
```diff
@@ -212,7 +212,18 @@ const config: HardhatUserConfig = {
   solidity: {
     compilers: [
       UNISWAP_SETTING,
-    ]
+    ],
+    overrides: {
```
KMKoushik, line 216: Timelock related contracts need `0.8.10` as it's used by compound

packages/hardhat/test/integration-tests/crabv2/crab-timelock.ts
```diff
@@ -0,0 +1,209 @@
...
+      currentBlockNumber = await provider.getBlockNumber()
+      const currentBlock = await provider.getBlock(currentBlockNumber)
+      eta = currentBlock.timestamp + twoDays + 300
+      currentBlockTimestamp = currentBlock.timestamp
+
+      data = abi.encode(['address'], [depositor2.address]);
+
+      txHash = ethers.utils.keccak256(abi.encode(['address', 'uint256', 'string', 'bytes', 'uint256'],
+        [crabStrategy.address, '0', signature, data, eta]))
+      await timelock.connect(owner).queueTransaction(crabStrategy.address, 0, signature, data, eta)
```
aparnakr, line 187: could you make this a test instead of in the before all and check that it doesn't work if we accidentally pass in bad data?

aparnakr (approved):
Are the tests also mostly from Compound? It might be fine to just port over and edit their tests.

aleone (approved):
lgtm

KMKoushik:
> Are the tests also mostly from Compound? It might be fine to just port over and edit their tests. 

That's what I did almost

KMKoushik:
> Is the contract size below the limit now? 

Yep

KMKoushik:
Will be merged in #530

PR #525, opened 2022-06-16
Otchedge v2 WIP

sanandnarayan:
This is Work in Progress.

Todo

- [ ] 1. Remove broken tests
- [ ] 2. Add manager check for time and price
- [ ] 3. Reusable hedge logic
- [ ] 4. spreadsheet version of rebalance logic
- [ ] 5. Tests for otc hedge

PR #533, opened 2022-06-21
Otchedgev2 rebased

sanandnarayan:
#525 rebased on #517


packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -37,9 +45,23 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
+
     /// @dev twap period to use for hedge calculations
     uint32 public hedgingTwapPeriod = 420 seconds;
 
+    /// @dev typehash for signed orders
+    bytes32 private constant _CRAB_BALANCE_TYPEHASH =
+        keccak256(
+            "Order(uint256 bidId,address trader,address traderToken,uint256 traderAmount,address managerToken,uint256 managerAmount,uint256 nonce)"
+        );
+
```
aleone, line 64: can kill the FLASH_HEDGE_SELL and FLASH_HEDGE_BUY in enum FLASH_SOURCE

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -91,54 +117,39 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
         uint256 minEth;
```
aleone, line 117: can kill FlashHedgeData

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -323,95 +361,6 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
-        bool _isStrategySellingWSqueeth,
-        uint256 _limitPrice
-    ) external payable nonReentrant {
-        require(_isPriceHedge(_auctionTriggerTime), "Price hedging not allowed");
-        _hedge(_auctionTriggerTime, _isStrategySellingWSqueeth, _limitPrice);
-
-        emit PriceHedge(msg.sender, _isStrategySellingWSqueeth, _limitPrice, _auctionTriggerTime);
-    }
-
-    /**
```
aleone, line 397: we might want re-add these as view functions, so our quoting bot or some f/e or back end can call it to check?

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -469,80 +418,16 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
 
-        emit SetMaxPriceMultiplier(_maxPriceMultiplier);
-    }
-
-    /**
-     * @notice get current auction details
-     * @param _auctionTriggerTime timestamp where auction started
-     * @return if strategy is selling wSqueeth, wSqueeth amount to auction, ETH proceeds, auction price, if auction direction has switched
-     */
-    function getAuctionDetails(uint256 _auctionTriggerTime)
```
aleone, line 514: Should we add a view function that tells the manager how much to hedge, maybe based on % of stratey to hedge or something.

Ie howMuchTohedge(50%) returns (true, 1000) (isSellingOSqth, quantity)

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -722,188 +572,156 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
+        bytes32 structHash = keccak256(
+            abi.encode(
+                _CRAB_BALANCE_TYPEHASH,
+                _order.bidId,
+                _order.trader,
+                _order.traderToken,
+                _order.traderAmount,
+                _order.managerToken,
+                _order.managerAmount,
+                _useNonce(_order.trader)
```
aleone, line 624: We don't do any nonce validation or even use it right? Shouldn't we check that the nonce provided is <= to current nonce?

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -722,188 +572,156 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
-     * @dev if _isHedgingOnUniswap, wSqueeth minted is kept to repay flashswap, otherwise sent to seller
-     * @param _buyer buyer address
-     * @param _buyerAmount buyer ETH amount sent
-     * @param _wSqueethToSell wSqueeth amount to sell
-     * @param _ethToBuy ETH amount to buy
-     * @param _isHedgingOnUniswap true if arbing with uniswap price
+     * @dev check the signer and swap tokens in the order
+     * @param managerSellAmount quantity the manager wants to sell
+     * @param managerBuyPrice the price at which the manager is buying
+     * @param _order a signed order to swap tokens
```
aleone, line 606: missing sellerPrice param

IMO lets use _sellerPrice (and similar _ for others) for consistency with internal variables.

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -722,188 +572,156 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
-            strategyDebt,
-            ethDelta,
-            auctionWSqueethEthPrice,
-            feeAdjustment
-        );
+    function hedgeOTC(
+        uint256 managerSellAmount,
+        uint256 managerBuyPrice,
+        Order[] memory _orders
+    ) external onlyOwner {
```
aleone, line 676: do we want this to be a multisig or EOA? If EOA, we might want to have something other than owner, so some is under multisig and some under EOA.

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -722,188 +572,156 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
-        (bool isStillSellingAuction, uint256 wSqueethToAuction) = _checkAuctionType(
-            strategyDebt,
-            ethDelta,
-            auctionWSqueethEthPrice,
-            feeAdjustment
-        );
+    function hedgeOTC(
+        uint256 managerSellAmount,
+        uint256 managerBuyPrice,
+        Order[] memory _orders
```
aleone, line 675: think no _ here

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -722,188 +572,156 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
+        _checkOTCPrice(managerBuyPrice, _orders[0].managerToken);
 
-        require(isSellingAuction == isStillSellingAuction, "auction direction changed");
+        timeAtLastHedge = block.timestamp;
 
-        uint256 ethProceeds = wSqueethToAuction.wmul(auctionWSqueethEthPrice);
+        uint256 remainingAmount = managerSellAmount;
+        uint256 prevPrice = 0;
+        uint256 currentPrice = 0;
+        bytes memory tradePair;
```
aleone, line 686: why bytes here?

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -722,188 +572,156 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
+        bytes memory prevTradePair;
```
aleone, line 687: I think can save some gas by optimizing this trade pair check.

Can compare each tradePair to the order[0].tradePair

``` bytes memory firstTradePair = abi.encode(_orders[0].traderToken, _orders[0].managerToken);
```

then, just requiring that ``` keccak256(tradePair) == keccak256(firstTradePair), ````

Additionally, any reason to not sure the keccak256 instead of the abi bytes? ie

``` bytes memory firstTradePair = keccak256(abi.encode(_orders[0].traderToken, _orders[0].managerToken)); ```

That way you only have to hash the firstTradePair 1x.

File: packages/hardhat/contracts/strategy/CrabStrategyV2.sol [hunk 11]
```diff
@@ -722,188 +572,156 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
-            feeAdjustment
-        );
+    function hedgeOTC(
+        uint256 managerSellAmount,
+        uint256 managerBuyPrice,
+        Order[] memory _orders
+    ) external onlyOwner {
+        require(managerBuyPrice > 0, "Manager Price should be greater than 0");
+        require(_isTimeHedge() || _isPriceHedge(), "Time or Price is not within range");
+        _checkOTCPrice(managerBuyPrice, _orders[0].managerToken);
```
aleone, line 679: I don't really like how we do this wrt to how we do prices for the otc hedge.  It seems odd that one of our prices is always in ETH/oSQTH terms and other is in managerToken/traderToken units.

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -722,188 +572,156 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
+            currentPrice = _orders[i].managerAmount.mul(1e18).div(_orders[i].traderAmount);
+            require(currentPrice >= prevPrice, "Orders are not arranged properly");
+            if (i > 0) {
+                require(
+                    keccak256(tradePair) == keccak256(prevTradePair),
+                    "All orders must have the same buy/sell token."
+                );
+            }
+            prevPrice = currentPrice;
+            prevTradePair = tradePair;
```
aleone, line 699: can remove if do above

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -469,80 +418,16 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
-    }
-
-    /**
-     * @notice owner can set the max price multiplier in a percentage scaled by 1e18 (11e18 is 110%)
-     * @dev the max price multiplier is multiplied by the TWAP price to get the final auction price
-     * @param _maxPriceMultiplier the max price multiplier, a percentage, scaled by 1e18
-     */
-    function setMaxPriceMultiplier(uint256 _maxPriceMultiplier) external onlyOwner {
-        require(_maxPriceMultiplier > 1e18, "max price multiplier too low");
+    function setOTCPriceTolerance(uint256 _otcPriceTolerance) external onlyOwner {
```
aleone, line 424: this maybe should be onlyTimelock

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -722,188 +572,156 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
+            // if last param is false, transfer happens again
+            _mintWPowerPerp(_order.trader, _order.managerAmount, _order.traderAmount, true);
+            priceAtLastHedge = _order.traderAmount.mul(1e18).div(_order.managerAmount);
+        } else {
+            // oSQTH in, WETH out
+            _burnWPowerPerp(_order.trader, _order.traderAmount, _order.managerAmount, true);
+            //wrap it
+            IWETH9(weth).deposit{value: _order.managerAmount}();
+            // if last param is false, transfer happens again
+            priceAtLastHedge = _order.managerAmount.mul(1e18).div(_order.traderAmount);
```
aleone, line 653: I think this has to be swapped.

priceAtLastHedge is stored as ETH per 1 oSQTH in contract, and this is oSQTH per ETH, correct?

aleone (approved):
Generally lgtm - approving, but some changes think we need.

Couple overarching comments:
- might want some setters to be onlyTimelock
- Slightly concerned with precision on our orders, but haven't explored fully yet.  Just thiniing that sometimes we use prices to figure out quantities and sometimes we use fixed quantities (trader and manager) and there may be some lost precision somewhere in there.
- I wonder if market makers will need a delegation address, ie if they want to sign orders with a multisig but it isnt the trading address.  I think this is something that was added in paradigm co.
- Cancellation and/or order expirations is probably important to have.
- Don't love that in some places in contract we use ETH per 1 oSQTH prices and in other places our prices are determined based on managerAmount / traderAmount (which could be reversed).

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -722,188 +537,157 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
...
+            }
+            prevPrice = currentPrice;
+            prevTradePair = tradePair;
+
+            if (remainingAmount > _orders[i].managerAmount) {
+                _execOrder(remainingAmount, _clearingPrice, currentPrice, _orders[i]);
+                remainingAmount = remainingAmount.sub(_orders[i].managerAmount);
+            } else {
+                _execOrder(remainingAmount, _clearingPrice, currentPrice, _orders[i]);
+                break;
```
aleone, line 672: Might want a view function that validates an order and tells the MM what will happen to it?

packages/hardhat/test/integration-tests/crabv2/crab-liquidation-dust.ts
```diff
@@ -3,7 +3,7 @@ import { expect } from "chai";
 import { Contract, BigNumber, providers } from "ethers";
 import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/dist/src/signer-with-address";
 import BigNumberJs from 'bignumber.js'
-import { WETH9, MockErc20, Controller, Oracle, WPowerPerp, CrabStrategyV2, ISwapRouter } from "../../../typechain";
+import { WETH9, MockErc20, Controller, Oracle, WPowerPerp, CrabStrategyV2, ISwapRouter, MockTimelock } from "../../../typechain";
```
aleone, line 6: I think for this we should be using no Mocks, right?

aleone: Though it doesn't really matter in this case, but maybe just a good practice for integration tests.

packages/hardhat/test/integration-tests/crabv2/crab-timelock.ts
```diff
@@ -0,0 +1,177 @@
...
+    CrabStrategyV2,
+    ISwapRouter,
+    MockTimelock,
+    Timelock,
+} from "../../../typechain";
+import {
+    deployUniswapV3,
+    deploySqueethCoreContracts,
+    deployWETHAndDai,
+    addWethDaiLiquidity,
```
aleone, line 21: delete dis

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
sanandnarayan: we do validate it, useNonce is the current nonce of the trader. So a nonce can be used only once, and when it is used it is incremented.
If that is not the case, the signature will not match. i.e the signer will not be equal to the trader

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
sanandnarayan: frontend can calculate this right? based on the vault collateral and shortAmount. What is the advantage of a view function on the contract?

packages/hardhat/.prettierrc
```diff
@@ -1,5 +1,4 @@
 {
-  "plugins": ["prettier-plugin-solidity"],
```
aleone, line 2: why we delete this? does this kill the solidity specific formatting?

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
sanandnarayan: which one of our prices is in ETH/oSQTH. I thought we always price in managerAmount/traderAmount

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
aleone: Maybe just for convenience.  It could be helpful for f/e to calculate it as well as let someone else (MM) or manager check the quantities that we are using.

packages/hardhat/test/integration-tests/crabv2/crab-liquidation-dust.ts
sanandnarayan: @KMKoushik

packages/hardhat/.prettierrc
sanandnarayan: this was causing my local to break for some reason. I did not want to commit it. I will revert this

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
aleone: Hmm, yeah that is good, but 20% is massive, and I'd probably prefer to exit strategy vs have a manager that could trade w that much slippage.

packages/hardhat/test/integration-tests/crabv2/crab-timelock.ts
```diff
@@ -0,0 +1,177 @@
...
+import BigNumberJs from "bignumber.js";
+import {
+    WETH9,
+    MockErc20,
+    Controller,
+    Oracle,
+    WPowerPerp,
+    CrabStrategyV2,
+    ISwapRouter,
+    MockTimelock,
```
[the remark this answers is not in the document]
sanandnarayan: @KMKoushik

packages/hardhat/test/integration-tests/crabv2/crab-timelock.ts
sanandnarayan: @KMKoushik

packages/hardhat/test/integration-tests/crabv2/crab-hedging-otc.ts
```diff
@@ -0,0 +1,1058 @@
...
+describe("Crab V2 flashswap integration test: time based hedging", function () {
+    const startingEthPrice = 3000;
+    const startingEthPrice1e18 = BigNumber.from(startingEthPrice).mul(one); // 3000 * 1e18
+    const scaledStartingSqueethPrice1e18 = startingEthPrice1e18.mul(11).div(10).div(oracleScaleFactor); // 0.303 * 1e18
+    const scaledStartingSqueethPrice = (startingEthPrice * 1.1) / oracleScaleFactor.toNumber(); // 0.303
+
+    const hedgeTimeThreshold = 86400; // 24h
+    const hedgePriceThreshold = ethers.utils.parseUnits("0.01");
+    const auctionTime = 3600;
+    const minPriceMultiplier = ethers.utils.parseUnits("0.95");
```
aleone, line 30: can delete these next 3

packages/hardhat/.prettierrc
sanandnarayan: this line, causes prettier to break for @KMKoushik as well. Can we remove this plugin? what is its use?

aparnakr (approved):

PR #535, opened 2022-06-22
added code to delegate signing for trade; needs tests

sanandnarayan:


aleone (approved):
lgtm, once tests added

PR #555, opened 2022-06-28
Bug Fix: Remaining amount shoud be greater than the quantity

sanandnarayan:
When remaining amount is greater than or equal to zero, we run the loop one extra time for the last order with zero quantity as remaining. This is not necessary


alpinechicken (approved):
lgtm

PR #556, opened 2022-06-28
Replace transfers with safeTransfer in hedgeOTC function

sanandnarayan:
Replace transfers with safeTransfer in hedgeOTC function


alpinechicken (commented):
also in strategyFlash 459 and 480?

aparnakr (approved):

aleone (approved):
I am ok with merging, but after looking into it more, think that we are actually increasing gas costs for users with no improvement in security as we know that squeeth and weth are conforming erc-20s. But again, ok w it.

aleone:
I am ok with merging, but after looking into it more, think that we are actually increasing gas costs for users with no improvement in security as we know that squeeth and weth are conforming erc-20s. But again, ok w it.

sanandnarayan:
@aleone agree, closing this PR without merging

PR #562, opened 2022-07-06
Crabv2 Audit Fixes: L1

aparnakr:
# Task:

## Description

Fixes issue L1 in Sherlock Audit report


daryakaviani (approved):

KMKoushik (approved):

aleone (approved):
generally lgtm, but one edge case I was thinking about that is weird is that if we somehow have a situation where there is >=1 wei stuck in this contract, the crab vault can never be fully closed (as this contract will always have some exposure that is stuck), which means that if we got there, we may be stuck in a dust state preventing users from being able to withdraw (without someone sticking 6.9 ETH in contract and basically leaving it there)

the fix is to allow transfer of oSQTH that is above totalCrabV1SharesMigrated to the owner (and add that variable back in)

aleone:
> 

If someone sends 1 wei to the contract instead of depositing it.

PR #569, opened 2022-07-06
[Audit Fix] - [I-8] - Approve EULER_MAINNET in the constructor

aleone:
# Task: I-8 - Approve EULER_MAINNET in the constructor

## Description

Move approval of euler to spend weth to constructor


aparnakr (approved):

packages/hardhat/contracts/strategy/CrabMigration.sol
```diff
@@ -120,6 +120,7 @@ contract CrabMigration is Ownable {
         weth = WETH9(_weth);
         dToken = _dToken;
         wPowerPerp = crabV1.wPowerPerp();
+        WETH9(_weth).approve(_eulerMainnet, type(uint256).max);
```
daryakaviani, line 123: can we just use `weth` here instead of reinitializing `WETH9(_weth)`?

daryakaviani (approved):
lgtm

packages/hardhat/contracts/strategy/CrabMigration.sol
aleone: We can't use any variables that are in the constructor for anything. So we have to use the params passed in (why using _eulerMainnet as well instead of eulerMainnet)

PR #572, opened 2022-07-06
[Audit Fix] - [L-8] - Add zero address checks to CrabMigration.sol constructor

daryakaviani:
Added zero address check for contract constructor for constructor parameter addresses.


aleone (changes requested):
Can you add unit tests for the expected reverts - similar to what we do for crab-strategy-v2 in the unit tests?

daryakaviani:
> Can you add unit tests for the expected reverts - similar to what we do for crab-strategy-v2 in the unit tests?

Added!

aparnakr (approved):

aleone (changes requested):
Can you have a test for weth = address(0) and crabv1 = address(0) so we know that all inputs will revert w address(0).

alpinechicken (approved):
lgtm

daryakaviani:
> Can you have a test for weth = address(0) and crabv1 = address(0) so we know that all inputs will revert w address(0).

Updated. CrabV1 reverted, but weth did not. Should we also add this check in `CrabStrategyV2.sol` then since we have an unchecked weth address being set?

aleone (approved):
lgtm

daryakaviani:
note: don't merge until 574 is merged + subsequent edit is made

PR #573, opened 2022-07-06
[Audit Fix] - Remove afterInitialization from most functions

aleone:
# Task:Remove afterInitialization from most functions to save gas

## Description
- remove afterInitialization from most functions
- set strategy cap in the initialize() function, 
- setStrategyCap has afterInitialization modifier
- update unit, integration, and e2e tests


alpinechicken (approved):
lgtm

sanandnarayan (approved):
looks good to me

packages/hardhat/contracts/strategy/CrabMigration.sol
```diff
@@ -154,15 +158,14 @@ contract CrabMigration is Ownable {
...
-        bytes memory data;
         euler.deferLiquidityCheck(
             address(this),
             abi.encode(
                 FlashloanCallbackData({
                     caller: msg.sender,
                     amountToBorrow: amountEthToBorrow,
                     callSource: uint8(FLASH_SOURCE.BATCH_MIGRATE),
-                    callData: data
+                    callData: abi.encode(BatchMigrate({strategyCap: _strategyCap}))
```
aparnakr, line 168: why not just set the strategy cap to be the amount transferred in initialize? Do we need to pass this in?

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
```diff
@@ -199,11 +199,14 @@ contract CrabStrategyV2 is StrategyBase, StrategyFlashSwap, ReentrancyGuard, Own
         uint256 _wSqueethToMint,
         uint256 _crabSharesToMint,
         uint256 _timeAtLastHedge,
-        uint256 _priceAtLastHedge
+        uint256 _priceAtLastHedge,
+        uint256 _strategyCap
```
aparnakr, line 203: this doesn't need to be passed in as a param right? Can just be the msg.value?

packages/hardhat/contracts/strategy/CrabMigration.sol
aleone: My thinking is that we probably want to set the initial cap > amount to migrate so once we migrate the strategy is live, but it could be done in this way with a call to setStrategyCap() afterwards.

packages/hardhat/contracts/strategy/CrabStrategyV2.sol
aleone: comment above on this, lmk your thoughts

packages/hardhat/contracts/strategy/CrabMigration.sol
aparnakr: ok cool

PR #576, opened 2022-07-07
[Audit Fix] - Allow Squeeth pool fee tier to be specified for flashDeposit and flashWithdrawal

aleone:
# Task: Allow Squeeth pool fee tier to be specified for flashDeposit and flashWithdrawal
## Description
- add in fee params for flashDeposit and flashWithdrawal
- add in 2nd fee params for flashDepositERC20 and flashWithdrawERC20
- update flash call back to look up the pool it needs to send funds to and pass correct data
- update _getPool to be internal instead of private
- update tests

Might clean up tests a bit as I don't love our use of isSimilar but think good to go.


aparnakr (approved):

packages/hardhat/test/setup.ts
```diff
@@ -244,8 +245,8 @@ export const addSqueethLiquidity = async(
       token0,
       token1,
       fee: feeTier,
-      tickLower: -887220,// int24 min tick used when selecting full range
-      tickUpper: 887220,// int24 max tick used when selecting full range
+      tickLower: is3000Fee ? -887220: -887200,// int24 min tick used when selecting full range
+      tickUpper: is3000Fee ? 887220: 887200,// int24 max tick used when selecting full range
```
daryakaviani, line 249: does this only work for 3% and 1% pool ticks? what if we do different pools down the line?

daryakaviani (approved):
lgtm

packages/hardhat/test/setup.ts
aleone: yeah this was hard for me to figure out why I was getting mint failures but the tick has to be divisible by the tick spacing

so implication for front end is that each mint to a different pool requires different tick upper / tick lower limits for full range liquidity

PR #578, opened 2022-07-07
Slippage quote fix

daryakaviani:
Before: using fixed slippage amount and overriding when sending minimum amount out
Now: using user inputted slippage amount from get-go

### DIFF
--- a/packages/frontend/src/components/Trade/Long/index.tsx
+++ b/packages/frontend/src/components/Trade/Long/index.tsx
@@ -391,7 +391,7 @@ const OpenLong: React.FC<BuyProps> = ({ activeStep = 0, open }) => {
       console.log(e)
       setBuyLoading(false)
     }
-  }, [buyAndRefund, ethTradeAmount, resetEthTradeAmount, resetSqthTradeAmount, setTradeCompleted, setTradeSuccess, slippageAmount])
+  }, [buyAndRefund, ethTradeAmount, resetEthTradeAmount, resetSqthTradeAmount, setTradeCompleted, setTradeSuccess])
 
   return (
     <div id="open-long-card">
--- a/packages/frontend/src/state/squeethPool/hooks.ts
+++ b/packages/frontend/src/state/squeethPool/hooks.ts
@@ -147,6 +147,7 @@ export const useGetBuyQuoteForETH = () => {
   const address = useAtomValue(addressAtom)
   const wethToken = useAtomValue(wethTokenAtom)
   const squeethToken = useAtomValue(squeethTokenAtom)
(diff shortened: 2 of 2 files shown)


packages/frontend/src/components/Trade/Long/index.tsx
```diff
@@ -390,7 +391,7 @@ const OpenLong: React.FC<BuyProps> = ({ activeStep = 0, open }) => {
       console.log(e)
       setBuyLoading(false)
     }
-  }, [buyAndRefund, ethTradeAmount, resetEthTradeAmount, resetSqthTradeAmount, setTradeCompleted, setTradeSuccess])
+  }, [buyAndRefund, ethTradeAmount, resetEthTradeAmount, resetSqthTradeAmount, setTradeCompleted, setTradeSuccess, slippageAmount])
```
adam5909099, line 394: I think the slippage amount is an unnecessary dependency.

daryakaviani: because it's a param?

packages/frontend/src/state/squeethPool/hooks.ts
```diff
@@ -483,15 +489,17 @@ export const useAutoRoutedGetSellQuote = () => {
...
       const chainId = networkId as any as ChainId
       const router = new AlphaRouter({ chainId: chainId, provider: (provider as any) })
       const rawAmount = CurrencyAmount.fromRawAmount(squeethToken!, fromTokenAmount(squeethAmount, OSQUEETH_DECIMALS).toFixed(0))
       const route = await router.route(rawAmount, wethToken!, TradeType.EXACT_INPUT,  {
         recipient: address!,
-        slippageTolerance: new Percent(5, 100),
+        slippageTolerance: parseSlippageInput(slippageTolerance.toString()),
         deadline: Math.floor(Date.now()/1000 +1800)
       })
+      console.log("slippageTolerance", slippageTolerance.toString())
```
adam5909099, line 502: Can we remove the console log here?

adam5909099 (approved):
👍

packages/frontend/src/components/Trade/Long/index.tsx
adam5909099: slippageAmount isn't used in the callback for transact so it's not needed in the dependency array.

KMKoushik:
Can you rebase this with main, so it will be easy to review?

daryakaviani:
> Can you rebase this with main, so it will be easy to review?

Done

KMKoushik (approved):

PR #581, opened 2022-07-08
[Audit Fix] - Added flags for nonces

sanandnarayan:


aleone (commented):
Initial quick look over changes:

- Think we need a way to query the status of a nonce (prb needed for the front end to make sure that the nonce chosen hasn't been used)
- Think we need a way to allow a user to toggle a nonce to true (ie replacement for incrementNonce() ) so they can cancel an order if wanted
- Concerned about how the front end will know what nonces are free for a user w/o iteration. 
   - One idea I had was that the contract tracks the highest nonce used for an address
   - mapping(address->highestNonce)
   - each _useNonce also checks if the _nonce is > highestNonce and if so sets it as the new noince
   - the f/e starts at highestNonce+1 for new orders each time (which may leave gaps for lower nonces, but that is prb ok and gives us an uninterrupted set of free nonces
   -the highestNonce can also be broken on the f/e by a user picking a very high nonce, say maxUint256, but that seems like just a f/e helper thing vs something that prevents functionality, the alternative below still works
   - Another alternative is that we just pick a random nonce for every order in the uint256 space, which is likely to be free :) 
-  tagging @KMKoushik for f/e thoughts on how we will select / track nonces

sanandnarayan:
@aleone Frontend generates a random nonce, and check if its used. If used, tried another mapping. The mapping is a public mapping, so anyone should be able to see if a nonce is used

sanandnarayan:
@aleone added the function to set nonce true

aleone (approved):
lgtm

aparnakr (approved):

daryakaviani (approved):
lgtm

aparnakr (approved):

alpinechicken (approved):
lgtm

PR #589, opened 2022-07-11
Fix Migration Error Codes

daryakaviani:
# Task: Fix Migration Error Codes

## Description

- Crab migration error codes were mixed up due to merges
- Fixing tests and contract to use correct error codes

### DIFF
--- a/packages/hardhat/contracts/strategy/CrabMigration.sol
+++ b/packages/hardhat/contracts/strategy/CrabMigration.sol
@@ -22,7 +22,7 @@ import {StrategyMath} from "./base/StrategyMath.sol";
  * M1: Migration already happened
  * M2: Migration has not yet happened
  * M3: msg.sender is not Euler Mainnet Contract
- * M4: msg.sender cannot send ETH
+ * M4: Can only receive ETH from weth, crabv1, or crabv2 contract
  * M5: Can't withdraw more than you own
  * M6: Not enough ETH to repay the loan
  * M7: Invalid crabV2 address
--- a/packages/hardhat/contracts/strategy/CrabMigration.sol
+++ b/packages/hardhat/contracts/strategy/CrabMigration.sol
@@ -19,18 +19,17 @@ import {StrategyMath} from "./base/StrategyMath.sol";
 
 /**
  * Migration Error Codes:
- * M1: Crab V2 Address already set
- * M2: Migration already happened
- * M3: Migration has not yet happened
- * M4: msg.sender is not Euler Mainnet Contract
- * M5: msg. sender cannot send ETH
- * M6: Can't withdraw more than you own
- * M7: Not enough ETH to repay the loan
- * M8: _ethToBorrow or _withdrawMaxEthToPay can't be 0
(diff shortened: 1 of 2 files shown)


aleone (approved):
lgtm

packages/hardhat/contracts/strategy/CrabMigration.sol
```diff
@@ -19,18 +19,17 @@ import {StrategyMath} from "./base/StrategyMath.sol";
...
- * M2: Migration already happened
- * M3: Migration has not yet happened
- * M4: msg.sender is not Euler Mainnet Contract
- * M5: msg. sender cannot send ETH
- * M6: Can't withdraw more than you own
- * M7: Not enough ETH to repay the loan
+ * M1: Migration already happened
+ * M2: Migration has not yet happened
+ * M3: msg.sender is not Euler Mainnet Contract
+ * M4: msg.sender cannot send ETH
```
alpinechicken, line 25: This is for the receive right?
`    receive() external payable {
        require(msg.sender == weth || msg.sender == crabV1 || msg.sender == crabV2, "M4");
    }`

aleone: Yeah it is. I think the meaning is that msg.sender (contract that is sending ETH) isnt allowed to, but maybe could be clearer.

packages/hardhat/contracts/strategy/CrabMigration.sol
```diff
@@ -19,18 +19,17 @@ import {StrategyMath} from "./base/StrategyMath.sol";
...
+ * M5: Can't withdraw more than you own
+ * M6: Not enough ETH to repay the loan
+ * M7: Invalid crabV2 address
```
alpinechicken, line 28: Mixed with M8? 
        require(_ethToBorrow > 0 && _withdrawMaxEthToPay > 0, "M7");

M8: _ethToBorrow or _withdrawMaxEthToPay can't be 0

daryakaviani: I believe the require statement is `require(_ethToBorrow > 0 && _withdrawMaxEthToPay > 0, "M8")` and tests also use M8?

packages/hardhat/contracts/strategy/CrabMigration.sol
daryakaviani: Agreed; now updated to "Can only receive ETH from weth, crabv1, or crabv2 contract".

packages/hardhat/contracts/strategy/CrabMigration.sol
alpinechicken: ah you're right - was looking at main

aleone: What line do you see ```require(_ethToBorrow > 0 && _withdrawMaxEthToPay > 0, "M7");```

I see that being M8 in the code and in the error legend.

alpinechicken (approved):
lgtm

PR #594, opened 2022-07-13
One-Click LP: Setup

daryakaviani:
# Task: One-Click LP Setup

## Description

Setup PR for addresses, ABIs, contract atoms, and price hooks. Learn more about the One-Click LP API [here](https://opynopyn.notion.site/One-Click-LP-ea85108a32ed40ff8c0c349a6728a1a4).


packages/frontend/src/state/controller/hooks.ts
```diff
@@ -210,6 +210,17 @@ export const useGetTwapEthPrice = () => {
...
 
+export const useGetTwapSqueethPrice = () => {
+  const { squeethPool, oSqueeth, weth } = useAtomValue(addressesAtom)
+  const { getTwapSafe } = useOracle()
+  const getTwapSqueethPrice = useCallback(
+    async () => await getTwapSafe(squeethPool, oSqueeth, weth, TWAP_PERIOD),
+    [oSqueeth, weth, getTwapSafe, squeethPool],
+  )
+
+  return getTwapSqueethPrice
```
adam5909099, line 221: Not a big deal but can we use useMemo instead and return twapPrice directly as sqeeuthPool, oSqueeth, weth won't change?
const twapSqueethPrice = useMemo(() => ...)
Just wanted to avoid performance issues due to calling getTwapSqueethPrice frequently

adam5909099: Sorry, useMemo can't be used here because getTwapSafe will return a different value depending on when it's called. There's a different solution for avoiding frequent API calls using useQuery.
https://github.com/opynfinance/squeeth-monorepo/pull/257/commits/c51e11460c9e5908a5feffc76c709b0d8d4971ec
But it shouldn't be used in LP-related hooks. So everything is good with the current implementation.

daryakaviani: Sounds good, will keep the current implementation. Why can't we use useQuery in LP-related hooks?

adam5909099: It uses cached data, and the cache will be updated every 10 seconds, so it's not perfectly accurate.

This hook is good enough for resolving this issue https://github.com/opynfinance/squeeth-monorepo/pull/257#issuecomment-1177025880 Alexis mentioned. The issue was related to fetching the twap price too frequently when the slider value changed, and this hook allows us to use previously fetched data assuming there won't be much change.

But I think LP required perfect accuracy and LP won't make frequent twap price calls so it doesn't seem to make sense to use stale data.

adam5909099 (approved):

packages/frontend/src/constants/address.ts
```diff
@@ -110,3 +110,10 @@ export const CRAB_STRATEGY: Address = {
 }
 
 export const ZERO_ADDR = '0x0000000000000000000000000000000000000000'
+
+export const CONTROLLER_HELPER: Address = {
```
alpinechicken, line 114: still need goerli deployed

packages/frontend/src/state/controller/hooks.ts
```diff
@@ -210,6 +210,17 @@ export const useGetTwapEthPrice = () => {
   return getTwapEthPrice
 }
 
+export const useGetTwapSqueethPrice = () => {
```
aleone, line 213: we probably want a way to get a slot0 or twap with period of 1 price, as that is more useful for things like LP compositions.

The twap should be used for things like vault CR.

aleone (approved):
smol comment, can be addressed later, but think will need other price functions for uniswap pools (not sure if exist elsewhere)

alpinechicken (approved):
lgtm

KMKoushik (approved):
LGTM

packages/frontend/src/state/controller/hooks.ts
KMKoushik: This is already available as an atom `squeethPriceeAtom`

PR #595, opened 2022-07-13
One-Click LP: Getter Hooks

daryakaviani:
# Task: One-Click LP Getter Hooks

## Description

Getter hooks that are to be used by one-click LP actions in future PRs:
- Getting position details from a uniswap token ID
- Getting position amounts from a static call to `decreaseLiquidity`
- Getting a quote for an oSQTH-ETH trade
- Getting the state of a pool from a Uniswap pool contract


packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,115 @@
...
+export const useGetPosition = () => {
+    const contract = useAtomValue(nftManagerContractAtom)
+  
+    const getPosition = useCallback(
+      async (uniTokenId: number) => {
+        if (!contract) return null
+        const position = await contract.methods.positions(uniTokenId).call()
+        const { nonce, operator, token0, token1, fee, tickLower, tickUpper, liquidity, feeGrowthInside0LastX128, feeGrowthInside1LastX128, tokensOwed0, tokensOwed1 } = position
+        return {
+          nonce,
```
adam5909099, line 27: It'd be great if we could define getting position using the useQuery hook from react-query with the key of ['position,' uniTokenId] so that we can take advantage of cache and get loading status, error message, etc.

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,115 @@
...
+  export const useGetDecreaseLiquidity = () => {
+    const contract = useAtomValue(nftManagerContractAtom)
+  
+    const getDecreaseLiquiduity = useCallback(
+      async (tokenId: number, liquidity: number, amount0Min: number, amount1Min: number, deadline: number) => {
+        if (!contract) return null
+        const DecreaseLiquidityParams = {
+          tokenId,
+          liquidity,
+          amount0Min,
```
adam5909099, line 56: I think we should use useDecreaseLiquidity instead of useGetDecreaseLiquidity to avoid confusion. Also, it'd be great if this hook can be based on useMutation so that we don't need to handle loading status and error manually.

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,115 @@
...
+/*** GETTERS ***/
+
+export const useGetPosition = () => {
+    const contract = useAtomValue(nftManagerContractAtom)
+  
+    const getPosition = useCallback(
+      async (uniTokenId: number) => {
+        if (!contract) return null
+        const position = await contract.methods.positions(uniTokenId).call()
+        const { nonce, operator, token0, token1, fee, tickLower, tickUpper, liquidity, feeGrowthInside0LastX128, feeGrowthInside1LastX128, tokensOwed0, tokensOwed1 } = position
```
adam5909099, line 25: @alexisgauba, @KMKoushik
How about adding a comment with the related doc on the Frontend code?
https://docs.uniswap.org/protocol/reference/periphery/NonfungiblePositionManager
I think we can improve the readability of the Frontend codebase if we have a habit of using basic JS Doc
https://jsdoc.app/about-getting-started.html

packages/frontend/src/state/lp/hooks.ts
adam5909099: Never mind, I realized in the meeting that I misunderstood the purpose of this function.

adam5909099 (approved):

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,115 @@
...
+      [contract],
+    )
+  
+    return getPosition
+  }
+
+  export const useGetDecreaseLiquidity = () => {
+    const contract = useAtomValue(nftManagerContractAtom)
+  
+    const getDecreaseLiquiduity = useCallback(
```
aleone, line 50: misspelling here ```getDecreaseLiquiduity```

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,115 @@
...
+        const DecreaseLiquidityParams = {
+          tokenId,
+          liquidity,
+          amount0Min,
+          amount1Min,
+          deadline,
+        }
+        console.log("DecreaseLiquidityParams", DecreaseLiquidityParams)
+  
+        const decreaseLiquidity = await contract.methods.decreaseLiquidity(DecreaseLiquidityParams).call()
```
aleone, line 62: what is this function used for/do?  is this called on the NonFungiblePositionManager contract?

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,115 @@
...
+  
+        return decreaseLiquidity
+      },
+      [contract],
+    )
+  
+    return getDecreaseLiquiduity
+  }
+
+  export const useGetQuote = () => {
```
aleone, line 72: mentioned in other PRs, but you probably need some different functions that do exactOut swaps in addition to exactIn swaps

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,115 @@
...
+  
+    const getQuote = useCallback(
+      async (amount: BigNumber, squeethIn: boolean) => {
+        if (!contract) return null
+  
+        const QuoteExactInputSingleParams = {
+          tokenIn: squeethIn ? oSqueeth : weth,
+          tokenOut: squeethIn ? weth : oSqueeth,
+          amountIn: amount.toFixed(0),
+          fee: 3000,
```
aleone, line 84: prb need to have this be adjustable in future, can be later though

packages/frontend/src/state/lp/hooks.ts
aleone: also slippage might be good to take into account here (for figuring out limit prices or figuring out the amountOutMax)

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,115 @@
...
+      },
+      [contract, weth, oSqueeth],
+    )
+  
+    return getQuote
+  }
+
+  async function getPoolState(squeethContract: Contract) {
+    const [slot, liquidity] = await Promise.all([
+      squeethContract?.methods.slot0().call(),
```
aleone, line 99: might want slot0 to price function

aleone (approved):
Most comments here are mentioned in other PRs so can be fixed/adjusted there.

packages/frontend/src/state/lp/hooks.ts
aleone: I guess I don't really undestand what is going on here. This is a write function right? Can we call it with a random address like this?  A user won't hold the NFT, so won't it revert?

daryakaviani: This is how we got the amount0 and amount1 in the ControllerHelper tests as well. It's just a static call (does not change contract state), which means it functions as view (not write).

There's other ways to get amount0 and amount1 with things like the `LiquidityAmounts.sol` contract, but this is the simplest.

packages/frontend/src/state/lp/hooks.ts
daryakaviani: Fixed

packages/frontend/src/state/lp/hooks.ts
daryakaviani: Is this a preferred way to get price instead of TWAP?

packages/frontend/src/state/lp/hooks.ts
alpinechicken: We could calculate from liquidty and ticks too but this is probably more precise

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,115 @@
...
+        const quote = await contract.methods.quoteExactInputSingle(QuoteExactInputSingleParams).call()
+        return quote.amountOut
+      },
+      [contract, weth, oSqueeth],
+    )
+  
+    return getQuote
+  }
+
+  async function getPoolState(squeethContract: Contract) {
```
alpinechicken, line 97: Maybe call poolContract to distinguish?

alpinechicken (approved):
lg

packages/frontend/src/state/lp/hooks.ts
KMKoushik: @daryakaviani Are we going to useQuery as Adam suggested?

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,147 @@
...
+
+      return decreaseLiquidity
+    },
+    [contract],
+  )
+
+  return getDecreaseLiquiduity
+}
+
+export const useGetExactIn = () => {
```
KMKoushik, line 77: @daryakaviani Seems like a duplicate. hooks in squeethPool has this functionality

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,147 @@
...
+      const quote = await contract.methods.quoteExactInputSingle(QuoteExactInputSingleParams).call()
+      return quote.amountOut
+    },
+    [contract, weth, oSqueeth],
+  )
+
+  return getExactIn
+}
+
+export const useGetExactOut = () => {
```
KMKoushik, line 102: Same. seems like duplicate

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,147 @@
...
+      const quote = await contract.methods.quoteExactOutputSingle(QuoteExactOutputSingleParams).call()
+      return quote.amountIn
+    },
+    [contract, weth, oSqueeth],
+  )
+
+  return getExactOut
+}
+
+async function getPoolState(poolContract: Contract) {
```
KMKoushik, line 127: Same comment. Seems like duplicate

KMKoushik (commented):
LGTM, but a lot's these functions are already available in squeethPool hooks

packages/frontend/src/state/lp/hooks.ts
daryakaviani: Got it - removing this one and adapting old one to include tickSpacing.

packages/frontend/src/state/lp/hooks.ts
daryakaviani: Few differences, so sticking with this approach:

- The squeethPool version was changed for the Auto Router for `exactIn`
- The squeethPool version doesn't allow for different fee tiers (going to add functionality for 1% pool next), so the `POOL_FEE` param will be customizable.
- Lots of token math in squeethPool version (converting to and from token amount several times to get the same value), so less elegant.

packages/frontend/src/state/lp/hooks.ts
daryakaviani: Same as previous comment.

packages/frontend/src/state/lp/hooks.ts
daryakaviani: I didn't add it since this function is just used in the other functions. Was thinking we can take advantage of the useQuery perks via the preexisting `useLPPositionsQuery` if we need to display user LP positions. Isn't this the case? @adam5909099 @KMKoushik 

If not, let me know if I need to make a different subgraph query to handle positions specifically generated by the API.

PR #596, opened 2022-07-13
One-Click LP: Open Position

daryakaviani:
# Task: One-Click LP Open Position

## Description

Open an LP position and deposit the LP NFT as collateral using a flashloan via the ControllerHelper's `flashloanWMintLpDepositNft` function.


packages/frontend/src/state/lp/hooks.ts
```diff
@@ -13,6 +13,61 @@ import { ethers } from 'ethers'
...
+  const handleTransaction = useHandleTransaction()
+  const getTwapSqueethPrice = useGetTwapSqueethPrice()
+  const getDebtAmount = useGetDebtAmount()
+  const openPositionDeposit = useAppCallback(
+    async (squeethToMint: BigNumber, lowerTickInput: number, upperTickInput: number, onTxConfirmed?: () => void) => {
+      if (!contract || !address) return null
+      
+      const squeethPrice = await getTwapSqueethPrice()
+      const mintWSqueethAmount = fromTokenAmount(squeethToMint, OSQUEETH_DECIMALS)
+      const ethDebt = await getDebtAmount(mintWSqueethAmount)
```
adam5909099, line 36: We can combine these promises using Promise.all.

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -13,6 +13,61 @@ import { ethers } from 'ethers'
...
+
+      const collateralToMint = ethDebt.multipliedBy(COLLAT_RATIO)
+      const collateralToLp = mintWSqueethAmount.multipliedBy(squeethPrice)
+
+      const lowerTick = nearestUsableTick(lowerTickInput, TICK_SPACE)
+      const upperTick = nearestUsableTick(upperTickInput, TICK_SPACE)
+
+      const flashloanWMintDepositNftParams = {
+        wPowerPerpPool: squeethPool,
+        vaultId: 0,
```
alpinechicken, line 46: always vaultId = 0?

adam5909099 (approved):

packages/frontend/src/state/lp/hooks.ts
daryakaviani: Ah, fixed!

packages/frontend/src/state/lp/hooks.ts
daryakaviani: Updated!

packages/frontend/src/state/lp/hooks.ts
aleone: i see this hard coded as vaultId: 0 still

aleone (changes requested):
Some comments on how we determine the params and some missing params

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -13,6 +13,62 @@ import { ethers } from 'ethers'
 import { useCallback } from 'react'
 import { useGetDebtAmount, useGetTwapSqueethPrice, useGetVault } from '../controller/hooks'
 
+/*** CONSTANTS ***/
+const TICK_SPACE = 60
+const COLLAT_RATIO = 1.5
```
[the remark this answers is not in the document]
daryakaviani: Yes we should let user specify. I think we should ship this, and then have a new PR that includes functionality for 1% pool with:
- Pool-specific tick amount
- Pool to use specified by user

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -13,6 +13,62 @@ import { ethers } from 'ethers'
...
+
+/*** ACTIONS ***/
+
+// Opening a mint and LP position and depositing
+export const useOpenPositionDeposit = () => {
+  const { squeethPool } = useAtomValue(addressesAtom)
```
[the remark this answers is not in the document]
daryakaviani: [Yes imo](https://github.com/opynfinance/squeeth-monorepo/pull/596/files#r935986580)

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -13,6 +13,62 @@ import { ethers } from 'ethers'
...
+
+      const lowerTick = nearestUsableTick(lowerTickInput, TICK_SPACE)
+      const upperTick = nearestUsableTick(upperTickInput, TICK_SPACE)
+
+      const flashloanWMintDepositNftParams = {
+        wPowerPerpPool: squeethPool,
+        vaultId: vaultId,
+        wPowerPerpAmount: mintWSqueethAmount.toFixed(0),
+        collateralToDeposit: collateralToMint.toFixed(0),
+        collateralToFlashloan: collateralToMint.toFixed(0),
```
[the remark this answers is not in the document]
daryakaviani: Fixed

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -13,6 +13,62 @@ import { ethers } from 'ethers'
 import { useCallback } from 'react'
 import { useGetDebtAmount, useGetTwapSqueethPrice, useGetVault } from '../controller/hooks'
 
+/*** CONSTANTS ***/
+const TICK_SPACE = 60
```
alpinechicken, line 17: can grab this with uniswapV3Pool.methods.tickSpacing().call()

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -13,6 +13,62 @@ import { ethers } from 'ethers'
...
+      if (!contract || !address) return null
+      
+      const mintWSqueethAmount = fromTokenAmount(squeethToMint, OSQUEETH_DECIMALS)
+      const squeethPricePromise = getTwapSqueethPrice()
+      const ethDebtPromise = getDebtAmount(mintWSqueethAmount)
+      const [squeethPrice, ethDebt] = await Promise.all([squeethPricePromise, ethDebtPromise])
+
+      const collateralToMint = ethDebt.multipliedBy(COLLAT_RATIO)
+      const collateralToLp = mintWSqueethAmount.multipliedBy(squeethPrice)
+
```
alpinechicken, line 41: Have example fns for amounts getAmounts, getNewAmounts at end here:  https://colab.research.google.com/drive/1oslr7uAN27q4-51iwY2X5lPqEDGTew32#scrollTo=1RF_rzrswNQL

but I think we got these in later PRs anyway

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -13,6 +13,62 @@ import { ethers } from 'ethers'
...
+        amount0Min: 0,
+        amount1Min: 0,
+        lowerTick: lowerTick,
+        upperTick: upperTick,
+      }
+
+      return handleTransaction(
+        contract.methods.flashloanWMintLpDepositNft(flashloanWMintDepositNftParams).send({
+          from: address,
+          value: collateralToLp.toFixed(0),
```
alpinechicken, line 62: won't just be sending collateralToLp in general right? Maybe collateralToLP + collateralToDeposit?

daryakaviani: I thought that these values overlap in the sense that we can LP using the squeeth that we mint? Is this not the case?

alpinechicken: We might be withdrawing eth from the vault to fund, or depositing extra eth into the vault

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -19,6 +19,78 @@ const POOL_FEE = 3000
...
+      const sqrtLowerPrice = new BigNumber(TickMath.getSqrtRatioAtTick(lowerTick).toString()).div(x96)
+      const sqrtUpperPrice = new BigNumber(TickMath.getSqrtRatioAtTick(upperTick).toString()).div(x96)
+      const squeethPrice = isWethToken0 ? new BigNumber(1).div(new BigNumber(TickMath.getSqrtRatioAtTick(Number(tick)).toString()).div(x96).pow(2))
+                                            : new BigNumber(TickMath.getSqrtRatioAtTick(Number(tick)).toString()).div(x96).pow(2)
+      const sqrtSqueethPrice = squeethPrice.sqrt()
+      const collateralToMint = ethDebt.multipliedBy(collatRatio)
+
+      // Lx = x * (sqrtSqueethPrice * sqrtUpperPrice) / (sqrtUpperPrice - sqrtSqueethPrice)
+      // y = Lx * (sqrtSqueethPrice - sqrtLowerPrice)
+      const liquidity = mintWSqueethAmount.times(sqrtSqueethPrice.times(sqrtUpperPrice)).div(sqrtUpperPrice.minus(sqrtSqueethPrice))
```
alpinechicken, line 54: I think we need a (p < pa) case for liquidity (we can exclude the case where oSQTH amount is zero). 

x = oSQTH amount
y = ETH amount

For liquidity L:

p < pa: L = x * (sqrtLowerPrice * sqrtUpperPrice) / (sqrtUpperPrice - sqrtLowerPrice)
pa < p < pb: L = x * (sqrtSqueethPrice * sqrtUpperPrice) / (sqrtUpperPrice - sqrtSqueethPrice)
(ignoring p>pb because then squeeth amount is zero)

Then collateralToLp below covers all the cases

alpinechicken: Maybe it's easiest to have all three cases for liquidity and collateralToLp and throw an error if p>pb

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -19,6 +19,78 @@ const POOL_FEE = 3000
...
+export const useOpenPositionDeposit = () => {
+  const { squeethPool } = useAtomValue(addressesAtom)
+  const address = useAtomValue(addressAtom)
+  const contract = useAtomValue(controllerHelperHelperContractAtom)
+  const handleTransaction = useHandleTransaction()
+  const getDebtAmount = useGetDebtAmount()
+  const squeethPoolContract = useAtomValue(squeethPoolContractAtom)
+  const isWethToken0 = useAtomValue(isWethToken0Atom)
+  const openPositionDeposit = useAppCallback(
+    async (squeethToMint: BigNumber, lowerTickInput: number, upperTickInput: number, vaultId: number, collatRatio: number, slippage: number, withdrawAmount: number, onTxConfirmed?: () => void) => {
```
KMKoushik, line 34: Not a blocker. Is there a doc on how to calculate these inputs? It will be beneficial for folks who build FE. Or else this will be just a wrapper around etherscan

daryakaviani: These inputs don't need to be calculated - they'll be directly inputted by the user! The most a calculation will be is dividing by 100 to get percentage to decimal format, etc.

The only substantial calculation left is converting between the squeethToMint param and if the user inputs an ETH equivalent to LP, but this will depend on UI decisions, so I can add that function later.

packages/frontend/src/state/lp/hooks.ts
daryakaviani: @alpinechicken Why can we exclude the case where oSQTH amount is 0? What if someone wants to open an all ETH position?

alpinechicken: maybe just because it would be a bit odd to just LP eth, but I guess there would be reasons. all three makes sense

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,219 @@
...
+      if (!contract || !address || !squeethPoolContract) return null
+      
+      const mintWSqueethAmount = fromTokenAmount(squeethToMint, OSQUEETH_DECIMALS)
```
alpinechicken, line 37: We should throw an error if squeethToMint is positive but our tick is in outside of the range and the vault needs to be all eth

packages/frontend/src/state/lp/hooks.ts
daryakaviani: I will make a doc for the "default" inputs once the design is finalized.

packages/frontend/src/state/lp/hooks.ts
alpinechicken: for p > pb:
L = y/(sqrt(pb)-sqrt(pa)

packages/frontend/src/state/lp/hooks.ts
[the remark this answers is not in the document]
aleone: ok so this will be added in future PR to support non 0.3% pools, that works for me

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,219 @@
...
+import useAppCallback from '@hooks/useAppCallback'
+import { addressAtom } from '../wallet/atoms'
+import { Contract } from 'web3-eth-contract'
+import { useHandleTransaction } from '../wallet/hooks'
+import { ethers } from 'ethers'
+import { useCallback } from 'react'
+import { useGetDebtAmount, useGetVault } from '../controller/hooks'
+
+/*** CONSTANTS ***/
+const COLLAT_RATIO_FLASHLOAN = 2
```
aleone, line 17: we aren't using this in the code, the flashloan eth needs to be determined using this

daryakaviani: we're using the user input for this in an upcoming commit - may use const in future hooks

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,219 @@
...
+      // Calculate prices from ticks
+      const [ethDebt, { tick, tickSpacing }] = await Promise.all([ethDebtPromise, poolStatePromise])
+      const lowerTick = nearestUsableTick(lowerTickInput, Number(tickSpacing))
+      const upperTick = nearestUsableTick(upperTickInput, Number(tickSpacing))
+      const sqrtLowerPrice = new BigNumber(TickMath.getSqrtRatioAtTick(lowerTick).toString()).div(x96)
+      const sqrtUpperPrice = new BigNumber(TickMath.getSqrtRatioAtTick(upperTick).toString()).div(x96)
+      const squeethPrice = isWethToken0 ? new BigNumber(1).div(new BigNumber(TickMath.getSqrtRatioAtTick(Number(tick)).toString()).div(x96).pow(2))
+                                            : new BigNumber(TickMath.getSqrtRatioAtTick(Number(tick)).toString()).div(x96).pow(2)
+      const sqrtSqueethPrice = squeethPrice.sqrt()
+      const collateralToMint = ethDebt.multipliedBy(collatRatio)
```
aleone, line 50: IMO this isn't how we should calculate CR - should match how f/e + contracts calculate it

We need to include the LP (ETH + oSQTH) in the calculation to solve for the ETH needed

```CR = (collateral+ethInLP+ethEquivalentOfoSQTH) / (oSQTHDebt * normFactor * ethPrice / 10000)```

File: packages/frontend/src/state/lp/hooks.ts (hunk 9 above)
Comment (aleone) [reply packages/frontend/src/state/lp/hooks.ts:54]: depending on isWethToken0, isn't it possible that 100% oSQTH is when we are below lowerTick instead of above upperTick?

Comment (aleone) [reply packages/frontend/src/state/lp/hooks.ts:54]: And yeah don't we have to calculate liquidity differently depending on what the oSQTH price is?

Comment (aleone) [reply packages/frontend/src/state/lp/hooks.ts:54]: (can get a negative liquidity for example if squeeth price is > upper tick price)

File: packages/frontend/src/state/lp/hooks.ts (hunk 5 above)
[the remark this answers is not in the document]
Comment (aleone) [reply packages/frontend/src/state/lp/hooks.ts:71]: collateralToDeposit and collateralToFlashloan shouldn't necessarily be the same

File: packages/frontend/src/state/lp/hooks.ts (hunk 13 above)
Comment (aleone) [reply packages/frontend/src/state/lp/hooks.ts:50]: Need to be 1 calc for how much eth to have in vault after the tx (collateralToDeposit) and 1 calc for how much eth to borrow (collateralToFlashloan)

File: packages/frontend/src/state/lp/hooks.ts [hunk 14]
```diff
@@ -0,0 +1,219 @@
...
+        amount0Min,
+        amount1Min,
+        lowerTick: lowerTick,
+        upperTick: upperTick,
+      }
+
+      return handleTransaction(
+        contract.methods.flashloanWMintLpDepositNft(flashloanWMintDepositNftParams).send({
+          from: address,
+          value: collateralToLp.plus(collateralToMint).minus(collateralToWithdraw).toFixed(0),
```
aleone, line 83: This can be negative, should handle that.

aleone (changes requested):
Think still some issues with how the logic is done for this, happy to discuss further.

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,243 @@
...
+      const vaultBefore = await getVault(vaultId)
+      if (!contract || !address || !squeethPoolContract || !vaultBefore || !vaultBefore.shortAmount || !vaultBefore.collateralAmount) return null
+      
+      const mintWSqueethAmount = fromTokenAmount(squeethToMint, OSQUEETH_DECIMALS)
+
+      // Calculate prices from ticks
+      const { tick, tickSpacing } = await getPoolState(squeethPoolContract)
+      const lowerTick = nearestUsableTick(lowerTickInput, Number(tickSpacing))
+      const upperTick = nearestUsableTick(upperTickInput, Number(tickSpacing))
+      const sqrtLowerPrice = new BigNumber(TickMath.getSqrtRatioAtTick(lowerTick).toString()).div(x96)
```
aleone, line 49: This tick section works only when weth is token 1.

We invert prices if weth is token 0, but in that case prices are represented by the ticks we choose as 10.xxx not 1/10.xxx.  We then compare 0.1xxx to 10.xx below.

We need to either invert all of them for comparison purposes or not invert them all.  The if else logic also depends on what we do.

If we keep prices in their native order and scrap the inversion (ie token1 per token0), then the if statement needs to be adjusted to account for which token is 0.

https://docs.google.com/spreadsheets/d/1ntfbm2WeXgPXJ0lI69RKABSbDx2LbbW5lxRXas0vRzY/edit?usp=sharing

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,243 @@
...
+
+      const collateralToWithdraw = fromTokenAmount(withdrawAmount, OSQUEETH_DECIMALS)
+      const ethIndexPrice = toTokenAmount(index, 18).sqrt()
+      const vaultShortAmt = fromTokenAmount(vaultBefore.shortAmount, OSQUEETH_DECIMALS)
+      const vaultCollateralAmt = fromTokenAmount(vaultBefore.collateralAmount, WETH_DECIMALS)
+      
+      // Calculate collateralToMint
+      const collateralToMint = (new BigNumber(collatRatio).times((vaultShortAmt.plus(mintWSqueethAmount)).times(normFactor).times(ethIndexPrice).div(INDEX_SCALE)))
+                              .minus((vaultCollateralAmt.minus(collateralToWithdraw).plus(collateralToLp).plus((mintWSqueethAmount).times(squeethPrice))))
+      const flashLoanAmount = (new BigNumber(collatRatio + FLASHLOAN_BUFFER).times(vaultShortAmt.plus(mintWSqueethAmount)).times(normFactor).times(ethIndexPrice).div(INDEX_SCALE))
```
aleone, line 85: This doesnt need to take into account collateralToMint

The flashloan only needs to collateralize short debt + new short debt at the set CR.

I think this should use the flashloan CR constant for collatRatio here.

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,243 @@
...
+      const amount0Min = amount0New.times(new BigNumber(1).minus(slippage)).toFixed(0)
+      const amount1Min = amount1New.times(new BigNumber(1).minus(slippage)).toFixed(0)
+
+      const collateralToWithdraw = fromTokenAmount(withdrawAmount, OSQUEETH_DECIMALS)
+      const ethIndexPrice = toTokenAmount(index, 18).sqrt()
+      const vaultShortAmt = fromTokenAmount(vaultBefore.shortAmount, OSQUEETH_DECIMALS)
+      const vaultCollateralAmt = fromTokenAmount(vaultBefore.collateralAmount, WETH_DECIMALS)
+      
+      // Calculate collateralToMint
+      const collateralToMint = (new BigNumber(collatRatio).times((vaultShortAmt.plus(mintWSqueethAmount)).times(normFactor).times(ethIndexPrice).div(INDEX_SCALE)))
```
aleone, line 83: We need to take into account the LP composition as our system sees it.

Our system treats oSQTH in LPs as an ETH amount equal to oSqthInLp * (EthPrice / 10000) * NF

So we need to add back the ETH in LP as well as the oSQTH value in ETH terms in the LP in this manner, not the eth value based on squeeth price.

Personally also got thrown off by the .minus with all the collateral together (some added) vs just a set of .minus(vaultCollateralAmt).plus(collateralToWithdraw).minus(collateralToLP)...etc, but it is right, just read it wrong.

aleone (changes requested):
some more fixes to LP math and the way collateral is calculated

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,243 @@
...
+      } else if (sqrtSqueethPrice.lt(sqrtLowerPrice)) {
+        // All squeeth position
+        collateralToLp = new BigNumber(0)
+      } else {
+        // Lx = x * (sqrtSqueethPrice * sqrtUpperPrice) / (sqrtUpperPrice - sqrtSqueethPrice)
+        // y = Lx * (sqrtSqueethPrice - sqrtLowerPrice)
+        const liquidity = mintWSqueethAmount.times(sqrtSqueethPrice.times(sqrtUpperPrice)).div(sqrtUpperPrice.minus(sqrtSqueethPrice))
+        collateralToLp = sqrtUpperPrice.lt(sqrtSqueethPrice) ? liquidity.times(sqrtUpperPrice.minus(sqrtLowerPrice))
+                              : sqrtSqueethPrice.lt(sqrtLowerPrice) ? new BigNumber(0)
+                              : liquidity.times(sqrtSqueethPrice.minus(sqrtLowerPrice))
```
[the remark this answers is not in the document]
daryakaviani: Conversion added to make everything token 1/token 0

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,243 @@
...
+        // All weth position
+        console.log("LPing an all WETH position is not enabled, but you can rebalance to this position.")
+        return
+      } else if (sqrtSqueethPrice.lt(sqrtLowerPrice)) {
+        // All squeeth position
+        collateralToLp = new BigNumber(0)
+      } else {
+        // Lx = x * (sqrtSqueethPrice * sqrtUpperPrice) / (sqrtUpperPrice - sqrtSqueethPrice)
+        // y = Lx * (sqrtSqueethPrice - sqrtLowerPrice)
+        const liquidity = mintWSqueethAmount.times(sqrtSqueethPrice.times(sqrtUpperPrice)).div(sqrtUpperPrice.minus(sqrtSqueethPrice))
```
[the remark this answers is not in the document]
daryakaviani: Conversion added to make everything token 1/token 0

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,243 @@
...
+      
+      const mintWSqueethAmount = fromTokenAmount(squeethToMint, OSQUEETH_DECIMALS)
+
+      // Calculate prices from ticks
+      const { tick, tickSpacing } = await getPoolState(squeethPoolContract)
+      const lowerTick = nearestUsableTick(lowerTickInput, Number(tickSpacing))
+      const upperTick = nearestUsableTick(upperTickInput, Number(tickSpacing))
+      const sqrtLowerPrice = new BigNumber(TickMath.getSqrtRatioAtTick(lowerTick).toString()).div(x96)
+      const sqrtUpperPrice = new BigNumber(TickMath.getSqrtRatioAtTick(upperTick).toString()).div(x96)
+      const squeethPrice = isWethToken0 ? new BigNumber(1).div(new BigNumber(TickMath.getSqrtRatioAtTick(Number(tick)).toString()).div(x96).pow(2))
```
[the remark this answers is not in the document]
daryakaviani: But I use squeethPrice itself as well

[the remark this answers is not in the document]
aleone: ah, I think you dont need to use it, see later comments

but fine to keep if you want

packages/frontend/src/state/lp/hooks.ts
daryakaviani: Fixed

aleone (approved):
Flashloan calc isn't 100% correct, see comment, fix that and good to merge.

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -0,0 +1,349 @@
...
+      const collateralToMint = new BigNumber(collatRatio)
+        .times(vaultShortAmt.plus(mintWSqueethAmount).times(normFactor).times(ethIndexPrice).div(INDEX_SCALE))
+        .minus(vaultCollateralAmt.minus(collateralToWithdraw).plus(collateralToLp).plus(oSQTHInETH))
+      const flashLoanAmount = new BigNumber(COLLAT_RATIO_FLASHLOAN + FLASHLOAN_BUFFER)
+        .times(vaultShortAmt.plus(mintWSqueethAmount))
+        .times(normFactor)
+        .times(ethIndexPrice)
+        .div(INDEX_SCALE)
+        .minus(vaultCollateralAmt.plus(collateralToMint).minus(collateralToWithdraw))
+      const collateralToMintPos = BigNumber.max(collateralToMint, 0)
```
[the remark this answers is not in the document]
daryakaviani: Sounds good, will note for future PR.

PR #597, opened 2022-07-13
One-Click LP: Close Position

daryakaviani:
# Task: One-Click LP Close Position

## Description

Close an LP position and withdraw the LP NFT that was used as collateral using a flashloan via the ControllerHelper's `flashloanCloseVaultLpNft` function.


packages/frontend/src/state/lp/hooks.ts
```diff
@@ -68,6 +69,61 @@ export const useOpenPositionDeposit = () => {
...
+      !position ||
+      !vaultBefore ||
+      !vaultBefore.shortAmount
+    )
+      return
+
+    const shortAmount = fromTokenAmount(vaultBefore.shortAmount, OSQUEETH_DECIMALS)
+    const debtInEth = await getDebtAmount(shortAmount)
+    const collateralToFlashloan = debtInEth.multipliedBy(COLLAT_RATIO)
+    const limitEth = await getQuote(shortAmount, true)
```
adam5909099, line 100: Can we combine these promises using Promise.all?

mindful-dev25 (approved):

packages/frontend/src/state/lp/hooks.ts
daryakaviani: Done!

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -68,6 +69,62 @@ export const useOpenPositionDeposit = () => {
   return openPositionDeposit
 }
 
+// Close position with flashloan
+export const useClosePosition = () => {
```
aleone, line 73: maybe naming could be improved (can have close positions w/o flashloans, etc)

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -68,6 +69,62 @@ export const useOpenPositionDeposit = () => {
...
+      const debtInEthPromise = getDebtAmount(shortAmount)
+      const limitEthPromise = getQuote(shortAmount, true)
+      const [debtInEth, limitEth] = await Promise.all([debtInEthPromise, limitEthPromise])
+      const collateralToFlashloan = debtInEth.multipliedBy(COLLAT_RATIO)
+
+    const flashloanCloseVaultLpNftParam = {
+      vaultId: vaultId,
+      tokenId: uniTokenId,
+      liquidity: position.liquidity,
+      liquidityPercentage: fromTokenAmount(1, 18).toFixed(0),
```
aleone, line 107: IMO for a more flexible API this should be flexible

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -68,6 +69,62 @@ export const useOpenPositionDeposit = () => {
...
+      wPowerPerpAmountToBurn: shortAmount.toFixed(0),
```
aleone, line 108: IMO for a more flexible API this should be flexible

packages/frontend/src/state/lp/hooks.ts
aleone: users could decide to close 50% of their LP to pay off 75% their debt as an example

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -68,6 +69,62 @@ export const useOpenPositionDeposit = () => {
...
+      !vaultBefore ||
+      !vaultBefore.shortAmount
+    )
+      return
+
+      const shortAmount = fromTokenAmount(vaultBefore.shortAmount, OSQUEETH_DECIMALS)
+      const debtInEthPromise = getDebtAmount(shortAmount)
+      const limitEthPromise = getQuote(shortAmount, true)
+      const [debtInEth, limitEth] = await Promise.all([debtInEthPromise, limitEthPromise])
+      const collateralToFlashloan = debtInEth.multipliedBy(COLLAT_RATIO)
```
aleone, line 101: needs to be higher, detailed comments on CollectFees PR

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -68,6 +69,62 @@ export const useOpenPositionDeposit = () => {
...
+
+    const flashloanCloseVaultLpNftParam = {
+      vaultId: vaultId,
+      tokenId: uniTokenId,
+      liquidity: position.liquidity,
+      liquidityPercentage: fromTokenAmount(1, 18).toFixed(0),
+      wPowerPerpAmountToBurn: shortAmount.toFixed(0),
+      collateralToFlashloan: collateralToFlashloan.toFixed(0),
+      collateralToWithdraw: 0,
```
aleone, line 110: IMO for a more flexible API this should be flexible

Most people that close 100% of an LP and burn 100% of debt will want to get the ETH out, the current construction requires them to do it in 2 txs.

IMO should be able to be set by the f /e

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -68,6 +69,62 @@ export const useOpenPositionDeposit = () => {
...
+      limitPriceEthPerPowerPerp: limitEth,
```
aleone, line 111: I think this param is incorrect and a couple of issues with how its being calculated.  

- This should be a limit price in units of ETHPerPowerPerp ie 0.1 ETH per 1 squeeth, but scaled by 1e18.
- The calculation is assuming that you are selling oSQTH, but this is not always the case, sometimes you have to buy, so you have to check if the debt being burned is > or < than the LP composition.
- Right now you are getting the amountOut from the swap (which is a reallllllly high limit price and will be sandwiched/always execute regardless of what market does)
- This also needs to use slippage somewhere a user should specify a slippage for this trade and that needs to be taken into account when the limit price is calculated.

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -68,6 +69,62 @@ export const useOpenPositionDeposit = () => {
...
+      amount0Min: 0,
+      amount1Min: 0,
+      poolFee: POOL_FEE,
+      burnExactRemoved: true,
```
aleone, line 115: burnExactRemoved means that none of the oSQTH will be bought or sold to burn the amount of debt listed (and the ```wPowerPerpAmountToBurn``` is not used), so there is no need for the above calculations and many of the params included are not used.

I think this is a good flag to expose in the API (most users may want to burnExactRemoved, but not all)

aleone: happy to discuss this and the above if any qs

aleone (changes requested):
Some changes to limit price as well as way the tx is generated / what its trying to do.

packages/frontend/src/state/controller/hooks.ts
```diff
@@ -210,6 +210,17 @@ export const useGetTwapEthPrice = () => {
   return getTwapEthPrice
 }
 
+export const useGetTwapSqueethPrice = () => {
```
alpinechicken, line 213: maybe + useGetSpotPrice using pool tick

daryakaviani: Yep, using this in the new approach [here](https://github.com/opynfinance/squeeth-monorepo/pull/521)!

packages/frontend/src/state/lp/hooks.ts
alpinechicken: flashClosePosition?

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -135,6 +136,130 @@ export const useOpenPositionDeposit = () => {
...
+
+      const flashLoanAmount = new BigNumber(COLLAT_RATIO_FLASHLOAN + FLASHLOAN_BUFFER)
+        .times(vaultShortAmt)
+        .times(normFactor)
+        .times(ethIndexPrice)
+        .div(INDEX_SCALE)
+        .minus(vaultCollateralAmt)
+      const flashLoanAmountPos = BigNumber.max(flashLoanAmount, 0)
+
+      const limitPrice = new BigNumber(limitEth)
```
alpinechicken, line 209: Think this should be (1+slippage) if we're buying oSQTH when amountToLiquidate > amountToBurn

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -135,6 +136,130 @@ export const useOpenPositionDeposit = () => {
...
+        .div(amountToLiquidate.minus(amountToBurn).abs())
+        .times(new BigNumber(1).minus(slippage))
```
alpinechicken, line 211: slippage for remove liquidity probably different to slippage for oSQTH/eth swap

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -135,6 +136,130 @@ export const useOpenPositionDeposit = () => {
...
+
+      const amountToLiquidate = new BigNumber(wPowerPerpAmountInLP).times(liquidityPercentage)
+      const amountToBurn = shortAmount.times(burnPercentage)
+      const limitEth = await (amountToLiquidate.gt(amountToBurn)
+        ? getExactIn(amountToLiquidate.minus(amountToBurn), true)
+        : getExactOut(amountToBurn.minus(amountToLiquidate), true))
+
+      const collateralToWithdraw = fromTokenAmount(withdrawAmount, WETH_DECIMALS)
+      const ethIndexPrice = toTokenAmount(index, WETH_DECIMALS).sqrt()
+      const vaultShortAmt = fromTokenAmount(vaultBefore.shortAmount, OSQUEETH_DECIMALS)
```
aleone, line 198: this is a repeat of shortAmount, we can delete

packages/frontend/src/state/lp/hooks.ts
aleone: yes, agreed

packages/frontend/src/state/lp/hooks.ts
aleone: yeah should probably be different variables, though for LP withdrawal slippage, it can probably not be surfaced for the user

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -135,6 +136,130 @@ export const useOpenPositionDeposit = () => {
...
+        0,
+        0,
+        Math.floor(Date.now() / 1000 + 86400),
+      )
+      const wPowerPerpAmountInLP = isWethToken0 ? amount1 : amount0
+
+      const amountToLiquidate = new BigNumber(wPowerPerpAmountInLP).times(liquidityPercentage)
+      const amountToBurn = shortAmount.times(burnPercentage)
+      const limitEth = await (amountToLiquidate.gt(amountToBurn)
+        ? getExactIn(amountToLiquidate.minus(amountToBurn), true)
```
aleone, line 193: right now, this doesn't take the pool fee in and do any autorouting, so we have to restrict this functionality to to be only the 0.3% pool

aleone (changes requested):
small changes needed

aleone (approved):
approving to not block, but small changes to be made

PR #598, opened 2022-07-13
One-Click LP: Collect Fees

daryakaviani:
# Task: One-Click LP Collect Fees

## Description

Collect fees accumulated by an LP position via the ControllerHelper's `rebalanceLpInVault` function.


mindful-dev25 (approved):

packages/frontend/src/state/lp/hooks.ts
```diff
@@ -124,6 +125,59 @@ export const useClosePosition = () => {
...
+      !address ||
+      !vaultBefore ||
+      !vaultBefore.shortAmount ||
+      !uniTokenId
+    )
+      return
+
+    const shortAmount = fromTokenAmount(vaultBefore.shortAmount, OSQUEETH_DECIMALS)
+    const debtInEth = await getDebtAmount(shortAmount)
+    const collateralToFlashloan = debtInEth.multipliedBy(COLLAT_RATIO)
```
aleone, line 152: for all of these we need to have some buffer room.

If the transaction takes some time to be mined and spot prices change, its possible we won't be borrowing enough ETH.

Maybe 200% is fine?

aleone (approved):
Small changes, big one that will impact all PRs is that we need to be more conservative with the amount of ETH that we borrow to target a higher CR.

approving to not block, but needs to be fixed

alpinechicken (approved):
lgtm

packages/frontend/src/constants/address.ts
```diff
@@ -124,3 +124,10 @@ export const CRAB_STRATEGY2: Address = {
 }
 
 export const ZERO_ADDR = '0x0000000000000000000000000000000000000000'
+
+export const CONTROLLER_HELPER: Address = {
+  1: '0xfa86d43b41Fa7a759c606130cc81970A955ff816',
+  3: '0x7e9C5490e91F93529c6480B46a59D738F6bcEa43',
```
alpinechicken, line 130: ropsten ded :(

alpinechicken: should add goerli deployment

aleone (approved):
lgtm, I think before merge we might want to remove all the other f/e code that doesnt deal with collect fees (ie useClosePosition, etc)

PR #603, opened 2022-07-13
Crabv2 deployment scripts

KMKoushik:
- Has deployment scripts
- Utils to create argument files


packages/hardhat/deploy/08_deploy_crab_v2.ts
```diff
@@ -0,0 +1,106 @@
...
+  await deploy("Timelock", {
+    from: deployer,
+    log: true,
+    args: timelockArgs,
+    skipIfAlreadyDeployed: true
+  });
+
+
+
+  console.log(`Successfully deploy Timelock`)
```
aparnakr, line 68: how do you get the addresses of the contracts deployed? Do you want to print that out in the script as well?

KMKoushik: It will automatically print

aparnakr (approved):

packages/hardhat/deploy/06_deploy_crab_strategy.ts
```diff
@@ -8,17 +8,20 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
   const wsqueeth = await ethers.getContract("WPowerPerp", deployer);
 
-  const {uniswapFactory} = await getUniswapDeployments(ethers, deployer, network.name)
+  const { uniswapFactory } = await getUniswapDeployments(ethers, deployer, network.name)
 
   const squeethPoolAddr = await getPoolAddress(wsqueeth, weth, uniswapFactory)
-  
-  if (network.name === "mainnet") {
+
+  if (network.name === "mainnet" || network.name === "ropsten") {
```
alpinechicken, line 24: doing this twice?

alpinechicken (approved):
lgtm - why won't it compile?

daryakaviani:
Crab V2 and ControllerHelper deployment circle tests do not work with hardhat because we do not have the proper addresses with hardhat. Changed circle tests to deploy using `npx hardhat deploy --ropsten` instead of `npx hardhat deploy --hardhat`.

PR #606, opened 2022-07-19
Added subgraph mappings for crabV2

sanandnarayan:
added subgraph mappings for crabV2


aparnakr (approved):

packages/subgraph/src/crabV2.ts
```diff
@@ -21,12 +29,85 @@ import {
...
+  userTx.user = event.params.withdrawer
+  userTx.owner = event.transaction.from
+  userTx.timestamp = event.block.timestamp
+  userTx.save()
+}
+
+export function handleFlashDeposit(event: FlashDeposit): void {
+  const userTx = loadOrCreateTx(event.transaction.hash.toHex())
+  userTx.wSqueethAmount = event.params.tradedAmountOut
+  userTx.ethAmount = (userTx.ethAmount !== null ? userTx.ethAmount : BigInt.fromString('0')).plus(event.transaction.value)
```
sanandnarayan, line 79: curious, when would userTx.ethAmount not equal null?

packages/subgraph/schema.graphql
```diff
@@ -89,3 +89,88 @@ type CrabAuction @entity {
...
+  id: ID!
+  vaultID: BigInt!
+  strategy: Bytes!
+  timestamp: BigInt!
+}
+
+type CrabUserTx @entity {
+  id: ID!
+  owner: Bytes!
+  user: Bytes!
```
sanandnarayan, line 171: @KMKoushik Should we create a user entity , and make all these tx's children of the user entity?
Also should we maintain a running account balances in that entity? Does it help for frontend or other analytics ; PnL?

KMKoushik: It would be a nice but not important IMO. We use on-chain balance to infer PnL to be sure!

packages/subgraph/src/crabV2.ts
KMKoushik: handleFlashDepositCallback  might be called before this. So it's safe to have this check

packages/subgraph/src/crabV2.ts
```diff
@@ -0,0 +1,203 @@
...
+  if(tx) {
+    tx.queued = false
+    tx.save();
+    execTimeLockTx.timelocktx = tx.id;
+  }
+  execTimeLockTx.save();
+}
+
+export function handleHedgeOTC(event: HedgeOTC): void {
+  const hedge = new HedgeOTCSchema(event.params.bidId.toString());
```
KMKoushik, line 148: Can we use tx here? Manager can give duplicate bidId here

packages/subgraph/src/crabV2.ts
```diff
@@ -0,0 +1,203 @@
...
+  hedge.bidID = event.params.bidId;
+  hedge.clearingPrice = event.params.clearingPrice;
+  hedge.quantity = event.params.quantity;
+  hedge.isBuying = event.params.isBuying;
+  hedge.timestamp = event.block.timestamp;
+  hedge.save();
+}
+
+export function handleHedgeOTCSingle(event: HedgeOTCSingle): void {
+  const hedge = new HedgeOTCSingleSchema(event.params.bidId.toString());
```
KMKoushik, line 158: Same! Also, every event for the same tx will have the same bid id and tx. So we need something more. I would recommend using tx-logId

KMKoushik (changes requested):

packages/subgraph/src/crabV2.ts
sanandnarayan: I need the ID to be bidID so that I can join then with HedgeOTCSingle

sanandnarayan: I think we might want to enforce uniqueness of bidID in frontend.

packages/subgraph/src/crabV2.ts
sanandnarayan: This done

packages/subgraph/src/crabV2.ts
sanandnarayan: Else we leave the join

KMKoushik (approved):

PR #609, opened 2022-07-22
Crab V2 Subgraph Transaction History: Hedge & User

daryakaviani:
- Added graphql queries for CrabV2 subgraph to frontend.
- Used queries to update hedge and user transaction history.

**V2 Hedge History:**
image

**V2 User Tx History:**
image


packages/frontend/src/hooks/useCrabPosition/useCrabPosition.ts
```diff
@@ -87,6 +87,84 @@ export const useCrabPosition = (user: string) => {
...
+
+  const index = useAtomValue(indexAtom)
+  const ethIndexPrice = toTokenAmount(index, 18).sqrt()
+
+  const [minCurrentEth, setMinCurrentEth] = useState(BIG_ZERO)
+  const [minCurrentUsd, setMinCurrentUsd] = useState(BIG_ZERO)
+  const [minPnlUsd, setMinPnlUsd] = useState(BIG_ZERO)
+  const [minPnL, setMinPnL] = useState(BIG_ZERO)
+
+  const { depositedEth, usdAmount: depositedUsd } = useAppMemo(() => {
```
KMKoushik, line 113: This seems good with one exception. Migrated users will not have a deposit tx. so this PnL / position might be wrong for users. We should track migration events too and consider them as deposits.  @alexisgauba What do you think will be the right thing here?

packages/frontend/src/components/Strategies/Crab/CapDetailsV2.tsx
```diff
@@ -0,0 +1,68 @@
+import CustomLinearProgress from '@components/CustomProgress'
```
KMKoushik, line 1: Can we create a new folder crabV2, so we clearly know the distinction?

KMKoushik (changes requested):
LGTM with some comments. Thanks for helping me out

aparnakr (approved):
The changes look fine to me, its working fine on the UI. 

The only thing I wonder is why its showing all the 27 files and changes which are already on the branch again. Have you rebased it?

KMKoushik:
Be merged #614

PR #619, opened 2022-07-25
Token Balance Bug Fix

daryakaviani:
Before, long and short positions only conditionally showed up since oSQTH balance showed up as 0.
https://user-images.githubusercontent.com/29926727/180891654-89bb6c88-b286-410c-9810-e19110c93ded.mov

This was due to an issue where polling was wrongfully set to false, even when variables only differed since one was null.


KMKoushik (approved):

adam5909099 (approved):
Looks good to me.

PR #625, opened 2022-07-27
Fix V2 positions

KMKoushik:


packages/frontend/src/hooks/contracts/useTokenBalance.ts
```diff
@@ -71,15 +73,11 @@ async function updateBalance(
   address: string | null,
   decimals: number,
 ) {
-  try {
```
daryakaviani, line 74: Why are we removing the try-catch here?

daryakaviani (approved):
lgtm - i am able to switch from mainnet to ropsten and see all my ropsten positions

daryakaviani:
Perhaps a separate issue, but when switching from Ropsten -> Mainnet, the Crab V2 position shows as loading even though I do not have a Crab V2 position on Mainnet:
image

alexisgauba (approved):

KMKoushik:
> Perhaps a separate issue, but when switching from Ropsten -> Mainnet, the Crab V2 position shows as loading even though I do not have a Crab V2 position on Mainnet: image

This should be fixed once the mainnet is migrated, if not we can come back to this

PR #627, opened 2022-07-28
Fox position for deposit v1 share tx 🦊 

KMKoushik:
Update proper eth amount for depsoit v1 tx using migration params


alexisgauba (approved):

daryakaviani (approved):
new math makes sense 2 meee

PR #628, opened 2022-07-28
feat: update profit thrshold for crab v2

alexisgauba:
# Task: Updates profit threshold for crab v2

## Description
Crab v1 had a profit threshold of 1 day. Update crab v2 to have a 2 day profit threshold (more conservative since it hedges between 2-3 days), and updates the copy to reflect profit threshold for between hedges vs. daily

Fixes # (issue)

## Type of change

- [ ] Document update

## How Has This Been Tested

Locally

## FE Checklist

## User Facing Checklist

- [x] I fully understand the user problem this PR is solving
- [x] I know who the target user is for this PR and have a deep understanding of that user
- [x] I have tried this flow thinking from the pov of the target user for this PR


aleone:
Looks good, a few comments:
v2:
- "if ETH moves less than approximately 6.12% in either direction before the next hedge"
  - this is really if ETH moves less than approximately 6.12% in either direction over a 2 day period
  - could also describe it as "if ETH moves less than approximately 6.12% between 2 day hedges"
  - might be worth stating that hedges may occur more frequently vs just MWF
  - same comment for the chart
- "Crab hedges approximately three times a week (on MWF), reducing risk risk of liquidations."
  - I don't love this, it implies that hedging 3 days per week is to reduce liquidations, but its really to get better prices, save $ for users
  - Could say that more frequent hedges could occur to reduce the risk of liquidation or prevent significant losses (ie we hedge in 1 day if the is a big move)
 
v1: the v1 copy is now changed to match v2 which isn't correct, can we have v1 be the old calculations/copy?

aleone (approved):
reviewed the preview, some comments above

packages/frontend/src/state/crab/utils.ts
```diff
@@ -35,8 +35,10 @@ export const getCollateralFromCrabAmount = async (
   return vault.collateralAmount.times(crabAmount).div(totalSupply)
 }
 
-export const getCurrentProfitableMovePercent = (currentImpliedFunding: number) => {
-  return Math.sqrt(currentImpliedFunding)
+export const getCurrentProfitableMovePercent = (currentImpliedVol: number) => {
+  // Approximating a hedge every 2 days, take the vol divided by the sqrt of # of periods
+  // In this case 365 / 2 = 182.5
+  return currentImpliedVol / Math.sqrt(182.5)
```
daryakaviani, line 41: I added a `YEAR` constant in the `constants.ts` file, which is 365. We can use that to avoid magic numbers.

daryakaviani (approved):
Approving, can update to use constant for 365 though!

alpinechicken (approved):
lgtm

PR #630, opened 2022-07-28
Fix tx confirmation showing early

KMKoushik:


alexisgauba (approved):

daryakaviani (commented):
image
I am seeing the revert for around 30s after deposit still.

KMKoushik:
> image
> 
> I am seeing the revert for around 30s after deposit still.

This seems to be some other issue can you check what's going on

daryakaviani:
> > image
> > I am seeing the revert for around 30s after deposit still.
> 
> This seems to be some other issue can you check what's going on

You just need to refresh/change tabs for it to work. Variables are probably still not initialized.

PR #632, opened 2022-08-01
WIP: Flashdeposit frontrun

KMKoushik:
- Sometimes `calculateETHtoBorrowFromUniswap` returns 0, because tick list length is 0 https://github.com/Uniswap/v3-sdk/blob/6e59ae3e893ae738767446c1e1d19a5c6a2c9588/src/utils/tickList.ts#L41
- In this fix I only initialize pool after getting the tick list data
- To make sure if this issue never happens, I am also showing an alert / throwing the error if ethToBorrow is 0

### DIFF
--- a/packages/frontend/src/state/crab/hooks.ts
+++ b/packages/frontend/src/state/crab/hooks.ts
@@ -238,7 +238,7 @@ export const useCurrentCrabPositionValue = () => {
   }, [userCrabBalance])
 
   useAppEffect(() => {
-    ;(async () => {
+    ; (async () => {
       setIsCrabPositionValueLoading(true)
       const [collateral, squeethDebt] = await Promise.all([
         getCollateralFromCrabAmount(userShares, contract, vault),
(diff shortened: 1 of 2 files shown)


packages/frontend/src/state/crab/hooks.ts
```diff
@@ -497,6 +497,11 @@ export const useFlashDepositV2 = (calculateETHtoBorrowFromUniswap: any) => {
       if (!contract || !vault) return
 
       let { ethBorrow: _ethBorrow } = await calculateETHtoBorrowFromUniswap(amount, slippage)
+      // Just to make sure the issue never happens
+      if (_ethBorrow.isZero()) {
+        alert('Some error occurred. Refresh the page!')
+        throw new Error('Some error occurred. Refresh the page!')
```
alexisgauba, line 503: Do we also want to disable the button in this case? Otherwise imo this works as an interim quick soln until the fix is complete

alexisgauba (approved):

PR #633, opened 2022-08-02
Fix flashdeposit bug

KMKoushik:
State is updated in a async function. Where it updates the state after calling getImmutables . This getImmutables can respond with any delay.

That's it should work normally. Where once you get all the ticks data (ticks.length = 547). It will be fully initialised pool.
Screen Shot 2022-08-02 at 3 37 08 pm

But due to async nature, rarely 2nd getImmutable be responded after the 3rd getImmutable. In that call ticks will be null(stale state) and you don't have the pool.
Screen Shot 2022-08-02 at 3 38 55 pm

### DIFF
--- a/packages/frontend/pages/index.tsx
+++ b/packages/frontend/pages/index.tsx
@@ -7,8 +7,9 @@ import InfoIcon from '@material-ui/icons/InfoOutlined'
 import KeyboardArrowDownIcon from '@material-ui/icons/KeyboardArrowDown'
 import KeyboardArrowUpIcon from '@material-ui/icons/KeyboardArrowUp'
 import Image from 'next/image'
-import { useState } from 'react'
+import { useCallback, useState } from 'react'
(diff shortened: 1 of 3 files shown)


alexisgauba (approved):

packages/frontend/pages/index.tsx
```diff
@@ -412,6 +413,11 @@ const SqueethInfo: React.FC = () => {
 
   const [showAdvanced, setShowAdvanced] = useState(false)
 
+  const toggleAdvanced = useCallback(() => {
+    setShowAdvanced((prevState) => !prevState)
+    Fathom.trackGoal('DRFFVOPH', 0)
```
daryakaviani, line 418: What is this acronym?

daryakaviani (approved):
code lgtm, but wasn't ever able to replicate the issue beforehand

PR #651, opened 2022-09-02
AML check update

DemolaJames:
# Task:

## Description

- Updated the api to return true for case where we do not get a 200 response for check provider 
- Added sentry logging for the checks that fall in the category above

Fixes # (issue)

## Type of change

- [ ] Document update

## How Has This Been Tested

## FE Checklist

## User Facing Checklist


packages/frontend/pages/api/isValidAddress.ts
```diff
@@ -18,8 +20,16 @@ const handleRequest = async (req: NextApiRequest, res: NextApiResponse) => {
     )
 
     res.status(200).json({ valid: (data?.[0]?.rating ?? 'highRisk') !== 'highRisk' })
```
akwritescode, line 22: what will happen, if you get a 404 to to the call in line 10?

DemolaJames: All errors not 200 goes Into the catch block...I tested for 400, 403, 404.

akwritescode (approved):
lgtm

packages/frontend/pages/api/isValidAddress.ts
KMKoushik, line 22: Did you try with risky address? To my knowledge, if you don't return at this point code will continue and always return as true

That brings my 2nd point, Can we add tests to make sure this always works? Can mock the `chainalysis` requests for different cases

KMKoushik (changes requested):

KMKoushik (approved):

PR #664, opened 2022-10-04
Crab OTC

sanandnarayan:


packages/hardhat/contracts/strategy/CrabOTC.sol
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * @dev Deposit into strategy by selling the minted oSqth
+     * @param _totalEth Total amount of ETH to deposit value + eth from selling minted oSqth
+     * @param _order A signed order to swap the tokens
+     */
+    function deposit(uint256 _totalEth, Order memory _order) external payable {
+        _verifyOrder(_order);
+        require(_order.isBuying, "Should be a buy order");
+
+        uint256 depositedEth = msg.value;
```
alpinechicken, line 88: Why are we depositing eth and then taking weth?

packages/hardhat/contracts/strategy/CrabOTC.sol
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /**
+     * @dev Withdraw from strategy by buying osqth
+     * @param _crabAmount Amount of crab to withdraw
+     * @param _order A signed order to swap the tokens
+     */
+    function withdraw(
+        uint256 _crabAmount,
+        uint256 _maxPrice,
```
alpinechicken, line 120: Seems unnecessary to have a maxPrice since initiator is coming in with the order which they presumably find acceptable?

alpinechicken (approved):
lgtm - I think _maxPrice is probably unnecessary for withdraw since the initiator is supplying the order from the mm, but makes sense as a double check. On a similar note _totalEth seems unnecessary except as a check for depositor since the deposit amount will be determined by msg.value and wethAmount.

PR #673, opened 2022-10-18
Feat: Init bull contracts

haythem96:
# Task: Init bull contracts

## Description

- BullStrategy.sol:
  - Inherit UniBull, LeverageBull
    - UniBull have all function that interact with Uni pool (I think we may need to move this later once we implement flashswap)
    - LeverageBull have all function that interact with Euler
  - Handle all strategy main features (deposit, rebalance, withdraw..etc)
  - it abstract how the depositor, withdrawer got access to funds (e.g for deposit, it assume sender have crab already, for withdraw, it assume sender have squeeth, usdc...etc)
- FlashBull:
  - Handle flash deposit, flashwithdraw, anything related to flashswaps
  - FlashBull do not inherit BullStrategy, it is a composition
- PayableBull:
  - Handle depositing into crab + sending crab into bull
  - PayableBull do not inherit BullStrategy, it is a composition

## Type of change

- [ ] Document update

## How Has This Been Tested

## FE Checklist

## User Facing Checklist


packages/bull-vault/src/BullStrategy.sol
```diff
@@ -0,0 +1,68 @@
...
+ * @notice BullStrategy contract
+ * @dev this is an abstracted BullStrategy in term of deposit and withdraw functionalities
+ * @author opyn team
+ */
+contract BullStrategy is ERC20, LeverageBull, UniBull {
+    using StrategyMath for uint256;
+
+    uint256 private constant ONE = 1e18;
+
+    address public immutable crab;
```
daryakaviani, line 21: We need to consider what if Crab changes and we want to set a new Crab contract address here, so probably not immutable.

packages/bull-vault/src/PayableBull.sol
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * @notice PayableBull contract
+ * @author opyn team
+ */
+contract PayableBull {
+    using StrategyMath for uint256;
+
+    BullStrategy private bullStrategy;
+
+    constructor(address _bull) {
```
daryakaviani, line 18: We should be specific about var names here since bull tokens, bull contract, etc., but I can just make an additional PR that sets our variable names.

daryakaviani (approved):
I don't think we should make the Crab address immutable, otherwise lgtm

haythem96:
> I don't think we should make the Crab address immutable, otherwise lgtm

Going to merge this for now, let's revisit this as if we will plan to change crab, we may need functions to migrate/withdraw from crab.

PR #680, opened 2022-10-21
[WIP] Bull Flashdeposit

daryakaviani:
Chaining flashswap approach

### DIFF
--- a/packages/bull-vault/src/BullStrategy.sol
+++ b/packages/bull-vault/src/BullStrategy.sol
@@ -102,7 +102,7 @@ contract BullStrategy is ERC20, LeverageBull, UniBull {
 
         uint256 ethUsdPrice = _getTwap(ethUSDCPool, weth, usdc, TWAP, false);
         uint256 squeethEthPrice = _getTwap(ethWSqueethPool, wPowerPerp, weth, TWAP, false);
-        (uint256 ethInCrab, uint256 squeethInCrab) = _getCrabVaultDetails();
+        (uint256 ethInCrab, uint256 squeethInCrab) = getCrabVaultDetails();
         uint256 crabUsdPrice = (ethInCrab.wmul(ethUsdPrice).sub(squeethInCrab.wmul(squeethEthPrice).wmul(ethUsdPrice)))
             .wdiv(IERC20(crab).totalSupply());
 
@@ -126,7 +126,7 @@ contract BullStrategy is ERC20, LeverageBull, UniBull {
         IERC20(_tokenIn).transfer(_pool, _amountToPay);
     }
 
-    function _getCrabVaultDetails() internal view returns (uint256, uint256) {
+    function getCrabVaultDetails() internal view returns (uint256, uint256) {
         VaultLib.Vault memory strategyVault = IController(powerTokenController).vaults(ICrabStrategyV2(crab).vaultId());
 
         return (strategyVault.collateralAmount, strategyVault.shortAmount);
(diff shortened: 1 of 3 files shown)


packages/bull-vault/src/FlashBull.sol
```diff
@@ -3,38 +3,163 @@ pragma solidity =0.7.6;
...
+     * @param _poolFee Uniswap pool fee
      */
-    function flashDeposit() external payable {}
+    function flashDeposit(uint256 _ethToCrab, uint256 _ethToDeposit, uint24 _poolFee) external payable {
+        uint256 ethUsdPrice = _getTwap(ethUSDCPool, weth, usdc, TWAP, false);
+        uint256 squeethEthPrice = _getTwap(ethWSqueethPool, wPowerPerp, weth, TWAP, false);
+        
+        (uint256 ethInCrab, uint256 squeethInCrab) = bullStrategy.getCrabVaultDetails();
+
+        // TODO: Helper function for Crab-USD price in BullStrategy.sol
```
haythem96, line 89: I think we should keep crab USD price separate instead of having this helper function as it will cost more to call

packages/bull-vault/src/LeverageBull.sol
```diff
@@ -59,13 +59,12 @@ contract LeverageBull {
      * @param _ethUsdPrice ETH price in USDC
      * @return ETH deposited as collateral in Euler and borrowed amount of USDC
      */
-    function _leverageDeposit(uint256 _crabAmount, uint256 _bullShare, uint256 _crabPrice, uint256 _ethUsdPrice)
-        internal
+    function calcLeverageEthUsdc(uint256 _crabAmount, uint256 _bullShare, uint256 _crabPrice, uint256 _ethUsdPrice)
```
haythem96, line 62: Would prefer keeping the code in the same block if we are not calling it in different places, easier to read and audit + less expensive as we don't need to call an internal function

packages/bull-vault/src/BullStrategy.sol
```diff
@@ -126,7 +126,7 @@ contract BullStrategy is ERC20, LeverageBull, UniBull {
         IERC20(_tokenIn).transfer(_pool, _amountToPay);
     }
 
-    function _getCrabVaultDetails() internal view returns (uint256, uint256) {
+    function getCrabVaultDetails() internal view returns (uint256, uint256) {
```
haythem96, line 129: Why removing the `_`, we usually write internal/private functions with `_` in the beginning

packages/bull-vault/src/LeverageBull.sol
daryakaviani: Was thinking about calling `calcLeverageEthUsdc` from `FlashBull`. Thoughts on best way to do that w/o repetitive code?

packages/bull-vault/src/BullStrategy.sol
daryakaviani: Was considering making it external func so I can use it in `FlashBull`. Should we just rewrite the function in `FlashBull`?

packages/bull-vault/src/LeverageBull.sol
haythem96: Also let's keep the functions in order, external -> public -> internal -> private (view->pure last) https://docs.soliditylang.org/en/v0.8.14/style-guide.html#order-of-functions

packages/bull-vault/src/BullStrategy.sol
haythem96: Let's keep `_getCrabVaultDetails()` internal so it can be called from contract that inherit BullStrategy and it will be cheaper, and have a separate `getCrabVaultDetails()` external that can be called from other contract where:
```
getCrabVaultDetails() {
  _getCrabVaultDetails()
}
```

daryakaviani:
Closing since we don't need nesting for flashswaps

PR #688, opened 2022-10-28
[WIP] Bull Auction: Leverage Rebalance

daryakaviani:
Bull auction skeleton & leverage rebalance


haythem96:
Did some refactoring, added owners and manager addresses, moved some state variables to the auction contract, should looks fine I guess.

haythem96:
Let's try to not work both on the same PR so we will not have biased review, so I think you can take it from there.

haythem96:
Also will be better to merge this with `fix/crab-balance-internal-state` branch and make this PR merge there also @daryakaviani

daryakaviani:
Updated to https://github.com/opynfinance/squeeth-monorepo/pull/699

PR #697, opened 2022-10-31
Second Flashdeposit Test

daryakaviani:
Added a test for 2 consecutive flashdeposits. Found an issue with calculations (potentially using wdiv(1e12) where unnecessary).


packages/bull-vault/test/integration-test/FlashBullTestFork.t.sol
```diff
@@ -153,6 +157,135 @@ contract FlashBullTestFork is Test {
...
+            ethToCrabInitial,
+            usdcToBorrow,
+            wSqueethToMint
+        );
+
+        vm.startPrank(user1);
+        flashBull.flashDeposit{value: totalEthToBull}(
+            ethToCrabInitial,
+            0,
+            0,
```
haythem96, line 198: I think we can set those to non-zero values, like getting the current prices from the pool and add a bit of slippage

packages/bull-vault/test/integration-test/FlashBullTestFork.t.sol
```diff
@@ -153,6 +157,135 @@ contract FlashBullTestFork is Test {
...
+
+        vm.startPrank(user1);
+        flashBull.flashDeposit{
+            value: calcTotalEthToBull(
+                wethToLendSecond,
+                ethToCrabSecond,
+                usdcToBorrowSecond,
+                wSqueethToMintSecond
+            )
+        }(ethToCrabSecond, 0, 0, 3000);
```
haythem96, line 265: Same here

packages/bull-vault/test/integration-test/FlashBullTestFork.t.sol
```diff
@@ -153,6 +157,135 @@ contract FlashBullTestFork is Test {
...
+                IEulerEToken(eToken)
+                    .balanceOfUnderlying(address(bullStrategy))
+                    .sub(wethToLendFirst)
+            ) <= 1
+        );
+        assertEq(
+            bullStrategy.getCrabBalance().sub(crabToBeMintedSecond),
+            bullCrabBalanceBeforeSecond,
+            "Bull crab balance mismatch for second flashdeposit"
+        );
```
haythem96, line 286: Maybe add assert for user bull balance

haythem96 (commented):
LGTM, added few comments.

packages/bull-vault/src/LeverageBull.sol
```diff
@@ -155,8 +155,9 @@ contract LeverageBull {
                 return (ethToLend, usdcToBorrow);
             }
         }
-        return (IEulerEToken(eToken).balanceOfUnderlying(address(this)).wmul(_bullShare).wdiv(ONE.sub(_bullShare)),
-               IEulerDToken(dToken).balanceOf(address(this)).wmul(_bullShare).wdiv(ONE.sub(_bullShare)).div(1e12));
+        uint256 ethToLend = IEulerEToken(eToken).balanceOfUnderlying(address(this)).wmul(_bullShare).wdiv(ONE.sub(_bullShare));
+        uint256 usdcToBorrow = IEulerDToken(dToken).balanceOf(address(this)).wmul(_bullShare).wdiv(ONE.sub(_bullShare));
```
haythem96, line 159: Yeah this make sense, IEulerDToken(dToken).balanceOf() is in 6 decimals already

packages/bull-vault/test/integration-test/FlashBullTestFork.t.sol
```diff
@@ -153,6 +157,135 @@ contract FlashBullTestFork is Test {
...
+        assertEq(
+            IEulerEToken(eToken).balanceOfUnderlying(address(bullStrategy)),
+            wethToLend
+        );
+        assertEq(
+            bullStrategy.getCrabBalance().sub(crabToBeMinted),
+            bullCrabBalanceBefore
+        );
+
```
haythem96, line 214: Add an assert for depositor bull balance check

packages/bull-vault/test/integration-test/FlashBullTestFork.t.sol
```diff
@@ -447,60 +630,136 @@ contract FlashBullTestFork is Test {
...
+        ) = _calcCollateralAndBorrowAmount(_crabToDeposit);
 
         IERC20(crabV2).approve(address(bullStrategy), _crabToDeposit);
         bullStrategy.deposit{value: wethToLend}(_crabToDeposit);
 
         return (wethToLend, usdcToBorrow);
     }
 
-    function _calcCollateralAndBorrowAmount(uint256 _crabToDeposit) internal view returns (uint256, uint256) {
+    function _calcCollateralAndBorrowAmount(uint256 _crabToDeposit)
```
haythem96, line 648: The extra 1e12 div is not fixed here

haythem96 (changes requested):
Need to remove the additional 1e12 div in test file

packages/bull-vault/test/integration-test/FlashBullTestFork.t.sol
daryakaviani: I fixed it in the else case I believe.

haythem96 (approved):
LGTM

packages/bull-vault/test/integration-test/FlashBullTestFork.t.sol
```diff
@@ -151,6 +196,197 @@ contract FlashBullTestFork is Test {
             bullStrategy.getCrabBalance().sub(crabToBeMinted),
             bullCrabBalanceBefore
         );
+        assertEq(
```
alpinechicken, line 199: Need to check user eth balance

alpinechicken: should be getting a small refund

packages/bull-vault/test/integration-test/FlashBullTestFork.t.sol
```diff
@@ -151,6 +196,197 @@ contract FlashBullTestFork is Test {
...
+            "Bull USDC debt amount mismatch for second flashdeposit"
+        );
+        assertTrue(
+            wethToLendSecond.sub(
+                IEulerEToken(eToken)
+                    .balanceOfUnderlying(address(bullStrategy))
+                    .sub(wethToLendFirst)
+            ) <= 1
+        );
+        assertEq(
```
alpinechicken, line 385: Add assert for user eth balance

packages/bull-vault/test/integration-test/FlashBullTestFork.t.sol
```diff
@@ -216,20 +467,31 @@ contract FlashBullTestFork is Test {
...
             "Bull ETH in leverage amount mismatch"
         );
-        assertEq(userBullBalanceBefore.sub(bullToRedeem), bullStrategy.balanceOf(user1), "User1 bull balance mismatch");
         assertEq(
-            crabBalanceBefore.sub(crabToRedeem), crabV2.balanceOf(address(bullStrategy)), "Bull crab balance mismatch"
+            userBullBalanceBefore.sub(bullToRedeem),
+            bullStrategy.balanceOf(user1),
+            "User1 bull balance mismatch"
+        );
+        assertEq(
```
alpinechicken, line 475: We should be able to get an exact value for this if we use uniswap quoter contract get exact eth out for squeeth and usdc. Same for flashdeposit.

packages/bull-vault/test/integration-test/FlashBullTestFork.t.sol
```diff
@@ -279,10 +573,19 @@ contract FlashBullTestFork is Test {
             bullCrabBalanceBefore,
             "Bull crab balance mismatch"
         );
+        assertEq(
+            bullToMint,
+            bullStrategy.balanceOf(user1),
+            "User1 bull balance mismatch"
+        );
```
alpinechicken, line 580: assert for user eth balance as above

alpinechicken (changes requested):
generally lgtm. I think we probably need an assert for user eth balance for flashWithdraw/flashDeposit because there will generally be refunds

PR #710, opened 2022-11-05
[WIP] Frontend for opening a new position with 1-click-lp

iamoperand:


packages/frontend/src/components/Lp/MintAndLp/DepositPreviewModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+      '& [class|="MuiStepConnector-line"]': {
+        minHeight: '40px',
+      },
+    },
+  }),
+)
+
+function getSteps() {
+  return ['Depositing ETH', 'Earning interest']
```
alexisgauba, line 64: lets actually just make this one step of 'Depositing ETH' since the user isn't earning fees + funding at their deposit, they earn fees + funding over time

packages/frontend/src/components/Lp/MintAndLp/PageHeader.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const classes = useStyles()
+
+  return (
+    <div className={classes.root}>
+      <div className={classes.container}>
+        <Typography variant="h1" className={classes.title}>
+          Deposit ERC-20s, earn ETH.
+        </Typography>
+        <Typography variant="subtitle1" className={classes.subtitle}>
+          Provide liquidity to earn interest through fees and funding.
```
alexisgauba, line 49:
```suggestion
          Provide liquidity to earn through fees.
```

packages/frontend/src/components/Lp/MintAndLp/PageHeader.tsx
```diff
@@ -0,0 +1,56 @@
...
+)
+
+const PageHeader: React.FC = () => {
+  const classes = useStyles()
+
+  return (
+    <div className={classes.root}>
+      <div className={classes.container}>
+        <Typography variant="h1" className={classes.title}>
+          Deposit ERC-20s, earn ETH.
```
alexisgauba, line 46:
```suggestion
          Deposit ETH, earn fees  
```

alexisgauba (changes requested):
Here liquidation price doesn't update as I adjust collat ratio - is that out of scope for this PR and being added in a diff PR? 

Also the default checkbox should uncheck itself if I adjust my collateralization ratio, and then if I re-check the box it should go back to the default ratio 

Screen Shot 2022-11-10 at 11 14 40 PM

We should also have the default price ranges filled out to be full range 

Screen Shot 2022-11-10 at 11 13 54 PM

packages/frontend/src/components/Lp/MintAndLp/DepositPreviewModal.tsx
iamoperand: Got it. If not "Earning interest", there should be something else I think. Otherwise the stepper design looks incomplete.

Screenshot 2022-11-21 at 7 16 45 PM

---

For now I'm removing the stepper component and just keeping "view dashboard" there.

Screenshot 2022-11-21 at 7 15 33 PM

packages/frontend/src/components/Lp/MintAndLp/PageHeader.tsx
iamoperand: done!

packages/frontend/src/components/Lp/MintAndLp/PageHeader.tsx
iamoperand: done!

PR #713, opened 2022-11-06
Feat: check cap at deposit

haythem96:
# Task: check cap at deposit

**Should be merged after #712**

## Description

check cap based on ETH deposited in Euler

## Type of change

- [ ] Document update

## How Has This Been Tested

## FE Checklist

## User Facing Checklist

### DIFF
--- a/packages/bull-vault/src/LeverageBull.sol
+++ b/packages/bull-vault/src/LeverageBull.sol
@@ -53,6 +53,7 @@ contract LeverageBull is Ownable {
 
     /**
      * @dev constructor
+     * @param _owner owner address
      * @param _euler euler address
      * @param _eulerMarkets euler markets module address
      * @param _powerTokenController wPowerPerp controller address
--- a/packages/bull-vault/src/LeverageBull.sol
+++ b/packages/bull-vault/src/LeverageBull.sol
@@ -57,7 +57,12 @@ contract LeverageBull is Ownable {
      * @param _eulerMarkets euler markets module address
      * @param _powerTokenController wPowerPerp controller address
      */
-    constructor(address _owner, address _euler, address _eulerMarkets, address _powerTokenController) Ownable() {
+    constructor(
+        address _owner,
+        address _euler,
+        address _eulerMarkets,
+        address _powerTokenController
+    ) Ownable() {
         eulerMarkets = _eulerMarkets;
         eToken = IEulerMarkets(_eulerMarkets).underlyingToEToken(
             IController(_powerTokenController).weth()
--- a/packages/bull-vault/src/LeverageBull.sol
+++ b/packages/bull-vault/src/LeverageBull.sol
@@ -8,6 +8,8 @@ import { IERC20 } from "openzeppelin/token/ERC20/IERC20.sol";
 import { IEulerMarkets } from "./interface/IEulerMarkets.sol";
 import { IEulerEToken } from "./interface/IEulerEToken.sol";
 import { IEulerDToken } from "./interface/IEulerDToken.sol";
+// contract
+import { Ownable } from "openzeppelin/access/Ownable.sol";
 // lib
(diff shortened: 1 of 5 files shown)


packages/bull-vault/src/LeverageBull.sol
```diff
@@ -124,7 +133,7 @@ contract LeverageBull {
         _depositEthInEuler(ethToLend, true);
         _borrowUsdcFromEuler(usdcToBorrow);
 
-        return (ethToLend, usdcToBorrow);
+        return (ethToLend, usdcToBorrow, IEulerEToken(eToken).balanceOfUnderlying(address(this)));
```
haythem96, line 136: Could be there an issue of using this balanceOf()? either here or for calc wethToLend overall? e.g if someone transfer some amount right before leverage or full rebalance ?

packages/bull-vault/test/integration-test/BullStrategyTestFork.t.sol
```diff
@@ -275,6 +298,10 @@ contract BullStrategyTestFork is Test {
             usdcToBorrow
         );
         assertEq(IERC20(usdc).balanceOf(user1).sub(userUsdcBalanceBefore), usdcToBorrow);
+        assertTrue(
```
aleone, line 301: if there ever a fuzz scenario that hits this assertion?  or is the cap too high relative to the crab balance user1 deposits in?

aleone: Can we test an explicit revert for the cap being hit?  Doesn't have to be in a fuzz test.

aleone (approved):
small comments, generally lgtm

think we should test the cap revert happening on deposit either through fuzz or explicitly in a test with params set to fail

packages/bull-vault/src/LeverageBull.sol
```diff
@@ -55,7 +57,12 @@ contract LeverageBull {
      * @param _eulerMarkets euler markets module address
      * @param _powerTokenController wPowerPerp controller address
      */
-    constructor(address _euler, address _eulerMarkets, address _powerTokenController) {
+    constructor(
+        address _owner,
+        address _euler,
+        address _eulerMarkets,
+        address _powerTokenController
+    ) Ownable() {
```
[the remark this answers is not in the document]
haythem96: Later we will need to have some functions in LeverageBull with onlyOwner modifier, like ones that will be called from Auction...etc

packages/bull-vault/src/LeverageBull.sol
```diff
@@ -105,7 +114,7 @@ contract LeverageBull {
      * @param _ethInCrab eth in crab strategy
      * @param _squeethInCrab oSQTH debt of crab strategy
      * @param _crabTotalSupply total supply of crab tokens
-     * @return ETH deposited as collateral in Euler and borrowed amount of USDC
+     * @return ETH deposited as collateral in Euler and borrowed amount of USDC, and total ETH deposited as collateral in Euler
```
alpinechicken, line 117: ETH deposited as collateral in Euler, and borrowed amount of USD, and total ETH deposited as collateral in Euler

packages/bull-vault/src/LeverageBull.sol
```diff
@@ -55,7 +57,12 @@ contract LeverageBull {
      * @param _eulerMarkets euler markets module address
      * @param _powerTokenController wPowerPerp controller address
      */
```
alpinechicken, line 59: add natspec for owner

alpinechicken (approved):
lgtm

packages/bull-vault/src/LeverageBull.sol
haythem96: added

haythem96:
> small comments, generally lgtm
> 
> think we should test the cap revert happening on deposit either through fuzz or explicitly in a test with params set to fail

Added

PR #725, opened 2022-11-12
Feat/ Improve BullStrategy coverage

aleone:
# Task:  Improve BullStrategy coverage with some missing tests

## Description
- call redeemCrabAndWithdrawWEth() with non-auction address, revert - Fixes ENG-868
- call depositEthIntoCrab() with non-auction address, revert  - Fixes ENG-869
- remove setCap 0 size restriction  - Fixes ENG-870
- tests successful cap set (assert correct value set)

Fixes ENG-867

PR coverage:
Screen Shot 2022-11-12 at 12 48 38 PM

Main coverage:
Screen Shot 2022-11-12 at 12 55 39 PM

### DIFF
--- a/packages/bull-vault/src/BullStrategy.sol
+++ b/packages/bull-vault/src/BullStrategy.sol
@@ -106,7 +106,6 @@ contract BullStrategy is ERC20, LeverageBull {
      * @param _cap strategy cap
      */
     function setCap(uint256 _cap) external onlyOwner {
-
         emit SetCap(strategyCap, _cap);
 
         strategyCap = _cap;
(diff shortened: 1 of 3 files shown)


alpinechicken (approved):

packages/bull-vault/src/BullStrategy.sol
```diff
@@ -104,11 +103,9 @@ contract BullStrategy is ERC20, LeverageBull {
 
     /**
      * @notice set strategy cap
-     * @param _cap startegy cap
+     * @param _cap strategy cap
      */
     function setCap(uint256 _cap) external onlyOwner {
-        require(_cap != 0, "BS1");
```
haythem96, line 110: Why removing this ?

haythem96: To use it as a ways to stop bull?

haythem96 (approved):

packages/bull-vault/src/BullStrategy.sol
aleone: yeah exactly

PR #772, opened 2022-12-02
fix: fix invalid input for long squeeth chart

iamoperand:
# Task:

"0" is not a valid input for historical days, but we still process it and get stuck. This PR fixes that.

### Before
Screenshot 2022-12-02 at 5 31 32 PM

### After
Screenshot 2022-12-02 at 5 31 53 PM

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Testing code


KMKoushik (approved):
LGTM

alexisgauba (changes requested):
zk's feedback was instead of showing an error message, don't let the user input 0 and instead start the input at 1 instead of 0

alexisgauba:
Updating the PR: next step here is to add start and end dates before we can merge per convo with @KMKoushik @iamoperand @zkoticha

iamoperand:
closing this as @rushikeshakhare has another PR for it: https://github.com/opynfinance/squeeth-monorepo/pull/798

PR #782, opened 2022-12-03
Fix: require for leverageDeposit

alpinechicken:
# Task:

## Description

leverageDeposit() takes _ethAmount input and reverts if not equal to wethToLend with LB0. This makes an initial deposit almost impossible because it requires user to know the eth/usd and eth/oSTH twaps ahead of time. 

This is mostly a problem for an initial deposit into bull, but can also be a problem for subsequent deposits if Euler e/d token balances are changing because this will the calculation of weth to lend.

Changes:
- Remove _ethAmount param for _leverageDeposit
- Change require to check  require(wethToLend <= msg.value, "LB0");
- Refund excess ETH 
- Tests for cases where msg.value is >/< wethToDeposit

Fixes ENG-1033

## Type of change

- [ ] New feature
- [ x] Bug fix
- [ ] Testing code


aleone (approved):
lgtm

haythem96 (approved):

PR #783, opened 2022-12-05
Add new Netting UI

KMKoushik:
Fixes ENG-910
Fixes ENG-923
Fixes ENG-938
Fixes ENG-937
Fixes ENG-933
Fixes ENG-932
Fixes ENG-946
Fixes ENG-945
Fixes ENG-942
Fixes ENG-941
Fixes ENG-966
Fixes ENG-955
Fixes ENG-965
Fixes ENG-964
Fixes ENG-950
Fixes ENG-943
Fixes ENG-929
Fixes ENG-954
Fixes ENG-952
Fixes ENG-1076
Fixes ENG-1075
Fixes ENG-1060
Fixes ENG-1079
Fixes ENG-1073


packages/frontend/src/components/Strategies/Crab/CrabPositionV2.tsx
```diff
@@ -82,14 +123,56 @@ const CrabPosition: React.FC = () => {
...
-              <Typography className={clsx(classes.metricValue, classes.white)}>
-                {formatCurrency(currentCrabPositionValue.toNumber())}
-              </Typography>
-            }
-          />
+          {usdcQueued.isGreaterThan(0) ? (
+            <Metric
+              label={
+                <MetricLabel
+                  label="Initiated Deposit"
```
alexisgauba, line 130: Looks like there is a loading error here: 
- When I first got to crab page it showed me no position 
- After I did a standard deposit of 300 USDC, it showed me a position of $261 https://goerli.etherscan.io/tx/0x554cf53f50d08de70ae332e664e1181099feffa42fb9a5f933bff06ae4d46a43
Screen Shot 2022-12-06 at 11 05 55 PM

- Was confused because initially thought it got my standard deposit wrong, but looks like there was an existing position from earlier 
- Went to withdraw to "instant" because I thought thats where I could dequeue since there was not "initaited deposit" with a cancel button, but that just flashWithdrew https://goerli.etherscan.io/tx/0xdd16feabb59084900670c0e08a9a08d6a2a47e5cda3883aac3f8b41f4473237c
- After the flashWithdraw it showed my queued deposit of 300 ETH with the "cancel button" but also my $261 position that should have been closed with the flashWithdraw

alexisgauba (changes requested):
On this tooltip, change the copy to `With standard deposit, your deposit has been initiated and will submitted via auction to avoid price impact. This may take up until Tuesday.` to make sure users know the auction refers to the standard deposit
Screen Shot 2022-12-06 at 11 24 06 PM

Couple of things here 
- should have the high price impact number be red, so its clear to users that high price impact is an issue. 
- we should show a high price impact warning 
- on errors like insufficient wallet balance, the button should become disabled instead of red with an outline
- Price impact and slippage should be on the same line, looks like a responsiveness issue when price impact hits double digits
Screen Shot 2022-12-05 at 11 34 40 AM

Says price impact is 1.61% when input says 0 so lets me do instant or standard -> looks like price impact is wrong here?

https://user-images.githubusercontent.com/13340486/206115549-1d615043-c8c3-4bd8-a69b-804b70357b67.mov

The get into the strategy by Tuesday tag is missing - never showed up for me on standard deposit 
Screen Shot 2022-12-06 at 11 03 57 PM




Couple of things on standard deposit 
- confirmation screen should say "initiated" instead of "queued" -> abstracting the concept of the queue for users
- button should just say "deposit", not "queue to avoid price impact" with the tooltip 
- the price impact should be the average auction price impact _not_ the Uniswap price impact, since on standard deposit they are going through the auction. For standard deposit we could also have a tooltip on price impact if we think this will be confusing, explaining that "For standard deposit, the average price impact is x% based on historical auctions"

packages/frontend/src/components/Strategies/Crab/CrabPositionV2.tsx
```diff
@@ -82,14 +123,56 @@ const CrabPosition: React.FC = () => {
...
+                  tooltipTitle="Your deposit has been initiated and will submitted via  auction to avoid price impact. This may take up until Tuesday"
```
alexisgauba, line 131:
```suggestion
                  tooltipTitle="Your deposit has been initiated and will be submitted via auction to avoid price impact. This may take up until Tuesday"
```

alexisgauba (changes requested):
Screen Shot 2022-12-12 at 7 03 57 PM
re price impact https://opyn.slack.com/archives/C04E8DQEZNC/p1670847369748499

PR #796, opened 2022-12-06
fix: don't divide by 0

iamoperand:
# Task:

Noticed that the value was amounting to NaN because of dividing by 0. This PR makes sure that doesn't happen.

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Testing code


KMKoushik (approved):
LGTM

alexisgauba (approved):

PR #814, opened 2022-12-09
fix: zen bull ui to match with crab

rushikeshakhare:
# Task:

Updated the UI to match Crab

## Description

- Update Zen Bull Graphic
- Font sizes for Bull Metrics
- Tooltips on Bull Metrics
- Load more with 3 and 3 incremental logic in strategy rebalances

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Testing code


alexisgauba (changes requested):
changes look good! this prob should be into a bull branch though right? shouldn't push to main before launch

KMKoushik (approved):

PR #817, opened 2022-12-11
Chore: fix rebase mess

haythem96:
# Task: fix rebase mess

## Description

- Remove repeated code due to wrong rebasing
- Remove unused vars in tests file and clean compile warnings

## Type of change

- [ ] New feature
- [ ] Bug fix
- [x] Testing code
- [ ] Document update or config files

PR #818, opened 2022-12-12
[Audit-Fix][L-03] improve L03 fix

haythem96:
# Task: improve L03 fix

**Should be merged after #817**

## Description

Improve L-03 fix

## Type of change

- [ ] New feature
- [X] Bug fix
- [ ] Testing code


packages/bull-vault/src/FlashBull.sol
```diff
@@ -67,16 +67,19 @@ contract FlashBull is UniFlash {
         uint256 ethToDepositInCrab;
     }
 
+    /// @dev struct for data encoding
```
aleone, line 70: I think this should maybe match the comment for `FlashDepositCrabData` as they are all the same between all of these? Or should be more specific and unique for each one.

aleone (approved):
I implemented this:

`The new docstrings (@notice message only) for LeverageBull::depositAndBorrowFromLeverage are incorrect since they correspond to the auctionRepayAndWithdrawFromLeverage function.
`

One small comment on the @dev for structs

alpinechicken (approved):
lgtm

PR #819, opened 2022-12-12
[Audit-Fix] Fix: farm()

haythem96:
# Task: fix farm()

**Should be merged after #818**

## Description

Restrict ETH farming during shutdown conditions

## Type of change

- [ ] New feature
- [X] Bug fix
- [ ] Testing code


packages/bull-vault/src/BullStrategy.sol
```diff
@@ -110,8 +111,10 @@ contract BullStrategy is ERC20, LeverageBull {
      * @param _receiver receiver address
      */
     function farm(address _asset, address _receiver) external onlyOwner {
+        require(!IController(powerTokenController).isShutDown(), "BS7");
```
aleone, line 114: I think technically these should only be in the `if (_asset == address(0)) {` section as we only want to restrict farming for ETH in that case.  For example, there could be euler tokens in the contract.  But I don't think it matters too much either way.

packages/bull-vault/src/BullStrategy.sol
```diff
@@ -110,8 +111,10 @@ contract BullStrategy is ERC20, LeverageBull {
...
+        require(!hasRedeemedInShutdown, "BS12");
```
aleone, line 115: same for this one

aleone (approved):
I think maybe we should only restrict ETH farming, but think either way is fine.

packages/bull-vault/src/BullStrategy.sol
alpinechicken: !hasRedeemedInShutDown => !isShutDown (contrapositive) so we only need one of these. I guess per the comment above this is to stop the owner farming out all the eth in shutdown and rugging users (is that right?). Makes sense if so, but only need to stop that specifically.

alpinechicken (approved):
lgtm. I think the first !isShutdown require is unnecessary and agree maybe still want to farm Euler tokens out in shutdown.

PR #825, opened 2022-12-12
[Audit Fix] ETHUSDC swaps have guardrails now

sanandnarayan:
prevent the rare case that there is some error in _p.minUSDC or _p.minETH to receive

Fixes https://github.com/sherlock-audit/2022-11-opyn-judging/issues/28


packages/crab-netting/src/CrabNetting.sol
```diff
@@ -490,6 +490,10 @@ contract CrabNetting is Ownable, EIP712 {
      */
     function depositAuction(DepositAuctionParams calldata _p) external onlyOwner {
         _checkOTCPrice(_p.clearingPrice, false);
+        uint256 ethUSDCPrice = IOracle(oracle).getTwap(ethUsdcPool, weth, usdc, auctionTwapPeriod, true);
+        require(
+            (_p.depositsQueued * (1e18 - MAX_OTC_PRICE_TOLERANCE) / ethUSDCPrice) < _p.minEth, "Min Eth Out too low"
```
aleone, line 495: this doesn't take into account decimals of USDC vs ETH

packages/crab-netting/src/CrabNetting.sol
```diff
@@ -672,13 +676,18 @@ contract CrabNetting is Ownable, EIP712 {
         }
 
         // step 4 convert to USDC
+        uint256 ethUSDCPrice = IOracle(oracle).getTwap(ethUsdcPool, weth, usdc, auctionTwapPeriod, true);
+        uint256 amountIn = (IERC20(weth).balanceOf(address(this)) - initWethBalance);
+        require(
+            (amountIn * ethUSDCPrice * (1e18 - MAX_OTC_PRICE_TOLERANCE) / 1e48) < _p.minUSDC, "Min USDC Out too low"
```
aleone, line 682: I don't love that it secretly adjusts for the USDC decimals in the 1e48 :)

aleone: Prefer explicit constants to do this adjustment

packages/crab-netting/src/CrabNetting.sol
aleone: it should also probably use `otcPriceTolerance` right not the constant?

packages/crab-netting/src/CrabNetting.sol
aleone: should use `otcPriceTolerance`, not the constant, right?

aleone (approved):
missing decimals on deposit(), small other comments

packages/crab-netting/src/CrabNetting.sol
sanandnarayan: Thanks for this catch

thec00n:
I think the checks for minimum swap out are ok. `((amountIn * ethUSDCPrice * (1e18 - otcPriceTolerance))` gets pretty close to uint max. For 10 ETH the result is 12737546894208079069871231751300000000000000000000000000000 and max uint is 115792089237316195423570985008687907853269984665640564039457584007913129639935. For this particular case where it uses, ETH amounts should be ok but if you use the lower value or higher decimal precision tokens then it will fail.

PR #846, opened 2022-12-15
Add Zen Bull vault license

haythem96:
# Task: Add Zen Bull vault license

## Description

Fixes ENG-1052

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Testing code
- [X] Document update or config files

### DIFF
--- a/packages/zen-bull-vault/src/FlashZen.sol
+++ b/packages/zen-bull-vault/src/FlashZen.sol
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: GPL-3.0-only
+//SPDX-License-Identifier: BUSL-1.1
 pragma solidity =0.7.6;
 
 pragma abicoder v2;
--- a/packages/zen-bull-vault/src/LeverageZen.sol
+++ b/packages/zen-bull-vault/src/LeverageZen.sol
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: GPL-3.0-only
+//SPDX-License-Identifier: BUSL-1.1
 pragma solidity =0.7.6;
 
 // interface
--- a/packages/zen-bull-vault/src/UniFlash.sol
+++ b/packages/zen-bull-vault/src/UniFlash.sol
@@ -1,5 +1,4 @@
-// SPDX-License-Identifier: GPL-2.0-or-later
-
+//SPDX-License-Identifier: BUSL-1.1
 pragma solidity =0.7.6;
 pragma abicoder v2;
 
--- a/packages/zen-bull-vault/src/UniOracle.sol
+++ b/packages/zen-bull-vault/src/UniOracle.sol
@@ -1,5 +1,4 @@
-// SPDX-License-Identifier: GPL-2.0-or-later
-
+//SPDX-License-Identifier: BUSL-1.1
 pragma solidity =0.7.6;
 pragma abicoder v2;
 
--- a/packages/zen-bull-vault/src/ZenAuction.sol
+++ b/packages/zen-bull-vault/src/ZenAuction.sol
@@ -1,6 +1,5 @@
-// SPDX-License-Identifier: GPL-3.0-only
+//SPDX-License-Identifier: BUSL-1.1
 pragma solidity =0.7.6;
-
 pragma abicoder v2;
 
 // interface
--- a/packages/zen-bull-vault/src/ZenBullStrategy.sol
+++ b/packages/zen-bull-vault/src/ZenBullStrategy.sol
@@ -1,6 +1,5 @@
-// SPDX-License-Identifier: GPL-3.0-only
+//SPDX-License-Identifier: BUSL-1.1
 pragma solidity =0.7.6;
-
 pragma abicoder v2;
 
 // interface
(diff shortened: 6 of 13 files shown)


packages/zen-bull-vault/src/interface/ICrabStrategyV2.sol
```diff
@@ -1,5 +1,4 @@
-// SPDX-License-Identifier: MIT
-
+//SPDX-License-Identifier: BUSL-1.1
```
aleone, line 1: imo all interfaces should be MIT

packages/zen-bull-vault/src/UniFlash.sol
```diff
@@ -1,5 +1,4 @@
-// SPDX-License-Identifier: GPL-2.0-or-later
-
+//SPDX-License-Identifier: BUSL-1.1
```
aleone, line 1: I don't think this can be BSL, I think it needs to be GPL due to all the uniswap imports, which are all GPL2

packages/zen-bull-vault/src/UniOracle.sol
```diff
@@ -1,5 +1,4 @@
-// SPDX-License-Identifier: GPL-2.0-or-later
-
+//SPDX-License-Identifier: BUSL-1.1
```
aleone, line 1: not sure this can be BSL, all the uniswap imports are GPL2

packages/zen-bull-vault/src/LeverageZen.sol
```diff
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: GPL-3.0-only
+//SPDX-License-Identifier: BUSL-1.1
```
aleone, line 1: ``StrategyMath.sol`` is imported here which is ``//SPDX-License-Identifier: AGPL-3.0-only
``, doesn't this mean we have to make it GPL?

aleone: It also imports our ``UniOracle.sol``, which I think needs to be licensed as GPL, but maybe thats ok?

packages/zen-bull-vault/src/FlashZen.sol
```diff
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: GPL-3.0-only
+//SPDX-License-Identifier: BUSL-1.1
```
aleone, line 1: ``StrategyMath.sol`` is imported here which is ``//SPDX-License-Identifier: AGPL-3.0-only ``, doesn't this mean we have to make it GPL?

It also imports our contract ``UniFlash.sol``, which I think needs to be GPL.  It also imports our contract ``VaultLib.sol, which is licensed as ``//SPDX-License-Identifier: GPL-2.0-or-later
``

packages/zen-bull-vault/src/UniOracle.sol
aleone: It also imports our contract ``OracleLibrary.sol`` which is ``// SPDX-License-Identifier: GPL-2.0-or-later
``, but maybe that is ok as its our contract?

packages/zen-bull-vault/src/ZenAuction.sol
```diff
@@ -1,6 +1,5 @@
-// SPDX-License-Identifier: GPL-3.0-only
+//SPDX-License-Identifier: BUSL-1.1
```
aleone, line 1: Imports our ``UniFlash.sol`` and ``UniOracle.sol`` which I think need to be GPL of some kind, but maybe its ok since our contract.

`StrategyMath.sol` is imported here which is `//SPDX-License-Identifier: AGPL-3.0-only` , doesn't this mean we have to make it GPL?

packages/zen-bull-vault/src/ZenBullStrategy.sol
```diff
@@ -1,6 +1,5 @@
-// SPDX-License-Identifier: GPL-3.0-only
+//SPDX-License-Identifier: BUSL-1.1
```
aleone, line 1: `StrategyMath.sol` is imported here which is `//SPDX-License-Identifier: AGPL-3.0-only` , doesn't this mean we have to make it GPL?

It also has our contracts `VaultLib.sol` and `LeverageZen.sol` which are GPL2, and maybe need to be GPL (for LeverageZen)

packages/zen-bull-vault/src/ZenEmergencyShutdown.sol
```diff
@@ -1,6 +1,5 @@
-// SPDX-License-Identifier: GPL-3.0-only
+//SPDX-License-Identifier: BUSL-1.1
```
aleone, line 1: `StrategyMath.sol` is imported here which is `//SPDX-License-Identifier: AGPL-3.0-only `, doesn't this mean we have to make it GPL?

We also import our contract `UniFlash.sol`, which may need to be GPL, see above.

aleone (changes requested):
@tamtam3000 I think we need your guidance here

packages/zen-bull-vault/src/interface/ICrabStrategyV2.sol
haythem96: why?

PR #868, opened 2022-12-21
fix: price impact warnings on crab & bull.

rushikeshakhare:
# Task:

Fix price impact warning for bull & crab when oSQTH value is trading lower than its minimum value. 

FIXES ENG-1124
FIXES ENG-1108

Note: we need to add more aggressive warning for this issue, right now fixing it with price impact warning. 

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Testing code

### DIFF
--- a/packages/frontend/src/components/Strategies/Bull/BullTrade/Deposit.tsx
+++ b/packages/frontend/src/components/Strategies/Bull/BullTrade/Deposit.tsx
@@ -163,15 +163,14 @@ const BullDeposit: React.FC<{ onTxnConfirm: (txn: BullTransactionConfirmation) =
     const squeethPrice = quote.ethOutForSqth.div(quote.oSqthIn)
     const scalingFactor = new BigNumber(INDEX_SCALE)
     const fundingPeriod = new BigNumber(FUNDING_PERIOD).div(YEAR)
-    const executionVol = new BigNumber(
-      Math.log(scalingFactor.times(squeethPrice).div(normFactor.times(ethIndexPrice)).toNumber()),
-    )
-      .div(fundingPeriod)
-      .sqrt()
-    const showPriceImpactWarning = executionVol
-      .minus(impliedVol)
-      .abs()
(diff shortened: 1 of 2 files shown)


KMKoushik:
Can you rebase this branch with main? I changed some logic in price impact calculation

rushikeshakhare:
Done

packages/frontend/src/components/Strategies/Bull/BullTrade/Deposit.tsx
```diff
@@ -163,15 +163,14 @@ const BullDeposit: React.FC<{ onTxnConfirm: (txn: BullTransactionConfirmation) =
...
-    )
-      .div(fundingPeriod)
-      .sqrt()
-    const showPriceImpactWarning = executionVol
-      .minus(impliedVol)
-      .abs()
-      .gt(BigNumber.max(new BigNumber(impliedVol).times(VOL_PERCENT_SCALAR), VOL_PERCENT_FIXED))
+    const log = Math.log(scalingFactor.times(squeethPrice).div(normFactor.times(ethIndexPrice)).toNumber())
+    const executionVol = new BigNumber(log).div(fundingPeriod).sqrt()
+    const showPriceImpactWarning =
```
aleone, line 168: I actually don't think we need this warning for the withdraw case for log<0 because this is a user buying oSQTH we are most worried about a user selling down to an undefined vol.  We still need it for the normal execution vol case (log >0 case)

If their execution vol is still NaN, that means they are buying when it is super cheap/NaN vol and its actually probably a good trade.

aleone (approved):
small comment, but other than that lgtm

packages/frontend/src/components/Strategies/Bull/BullTrade/Deposit.tsx
rushikeshakhare: This change is only being made for deposit cases.

KMKoushik (approved):

PR #871, opened 2022-12-23
fix: eth & squeeth prices

rushikeshakhare:
# Task:

Fetch Eth Price & squeeth price as twap every 15 seconds. 

Fixes ENG-1145

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Testing code


packages/frontend/src/hooks/useETHPrice.ts
```diff
@@ -37,8 +37,7 @@ export const useETHPrice = (refetchIntervalSec = 30) => {
  */
 export const useOnChainETHPrice = () => {
   const index = useAtomValue(indexAtom)
-
-  return toTokenAmount(index, 18).sqrt()
+  return index
```
KMKoushik, line 40: the name `useOnChainETHPrice ` implies it's going to return ethPrice, it's better to return eth price instead of sqrt()

packages/frontend/src/state/controller/hooks.ts
```diff
@@ -360,36 +359,30 @@ const useNormFactor = () => {
...
 
-  useEffect(() => {
-    if (!contract) return
-    getIndex(1, contract).then(setIndex)
-  }, [address, networkId, contract, setIndex])
+  useAppEffect(() => {
+    const getTwapEth = () => {
+      if (ethUsdcPool && weth && usdc) {
+        getTwapSafe(ethUsdcPool, weth, usdc, 1).then((quote) => {
+          setIndex(quote)
```
KMKoushik, line 371: It's better to use a new atom of ethPrice eg: `ethPriceAtom`. Index is actually taken from our contract and it serves separate purpose.

packages/frontend/src/state/controller/hooks.ts
```diff
@@ -360,36 +359,30 @@ const useNormFactor = () => {
...
-      'logs',
-      {
-        address: [ETH_USDC_POOL[networkId]],
-        topics: [SWAP_EVENT_TOPIC],
-      },
-      () => {
-        getIndex(1, contract).then(setIndex)
-      },
-    )
+    getTwapEth()
```
KMKoushik, line 376: Can we change this to a hook. so it will be easy to adapt for any other states. https://overreacted.io/making-setinterval-declarative-with-react-hooks/

KMKoushik (changes requested):
Overall good with some comments

KMKoushik (approved):

PR #878, opened 2022-12-28
feat: separate uniswap fee and price impact on crab

rushikeshakhare:
# Task:

Separate Uniswap fees & Price Impact on Deposit/Withdraw Crab

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Testing code


alexisgauba (changes requested):
Overall lgtm! 

Only issue is that here, its giving me a high price impact warning but not switching to standard / standard is still greyed out: 
Screen Shot 2022-12-28 at 4 00 36 PM

KMKoushik (changes requested):
Currently, price impact determines if we have to use standard deposit or not. This will break that. So `fee` should also be included in the calculation

KMKoushik (approved):

PR #886, opened 2022-12-29
fix: crab position issues after confirmed transaction

rushikeshakhare:
# Task:

This PR is ready as per the code fix it needed, but it still needs to be tested with new crab design.

Right now new crab does not work with Goerli 

FIXES ENG-1193
FIXES ENG-1205

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Testing code

### DIFF
--- a/packages/frontend/src/hooks/useCrabPosition/useCrabPosition.ts
+++ b/packages/frontend/src/hooks/useCrabPosition/useCrabPosition.ts
@@ -11,7 +11,6 @@ import useAppMemo from '../useAppMemo'
 import {
   crabLoadingAtom,
   crabLoadingAtomV2,
-  crabPositionRefetchingAtom,
   crabPositionValueLoadingAtom,
   crabPositionValueLoadingAtomV2,
   crabQueuedInEthAtom,
@@ -108,7 +107,7 @@ export const useCrabPosition = (user: string) => {
 */
 export const useCrabPositionV2 = (user: string) => {
   const crabLoading = useAtomValue(crabLoadingAtomV2)
-  const [crabPositionRefetching, setCrabPositionRefetching] = useAtom(crabPositionRefetchingAtom)
+  const [crabPositionRefetching, setCrabPositionRefetching] = useState(false)
   const isCrabPositionValueLoading = useAtomValue(crabPositionValueLoadingAtomV2)
   const currentEthValue = useAtomValue(currentCrabPositionValueInETHAtomV2)
   const [txToSearch, setTxToSearch] = useState<string | undefined>(undefined)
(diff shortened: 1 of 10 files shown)


packages/frontend/src/hooks/useCrabPosition/useCrabPosition.ts
```diff
@@ -107,11 +108,13 @@ export const useCrabPosition = (user: string) => {
 */
 export const useCrabPositionV2 = (user: string) => {
   const crabLoading = useAtomValue(crabLoadingAtomV2)
+  const [crabPositionRefetching, setCrabPositionRefetching] = useAtom(crabPositionRefetchingAtom)
```
iamoperand, line 111: can we just use `useState` here to keep things simple? as far as I can see this state is not being used elsewhere.

packages/frontend/src/hooks/useUserCrabV2TxHistory.ts
```diff
@@ -66,8 +66,13 @@ export const useUserCrabV2TxHistory = (user: string, isDescending?: boolean) =>
       })
   }, [crabUserTxes])
 
+  const isEthUsdPriceMapValid = useMemo(
+    () => ethUsdPriceMap && crabUserTxes.map((tx) => !!ethUsdPriceMap![Number(tx.timestamp) * 1000]),
```
iamoperand, line 70: instead of `map`, don't you think `every` makes more sense here?

packages/frontend/src/state/crab/hooks.ts
```diff
@@ -424,7 +424,11 @@ export const useCurrentCrabPositionValueV2 = () => {
     crabQueuedShares,
   ])
 
-  return { currentCrabPositionValue, currentCrabPositionValueInETH, isCrabPositionValueLoading }
+  return {
+    currentCrabPositionValue,
+    currentCrabPositionValueInETH,
+    isCrabPositionValueLoading: balLoading || isCrabPositionValueLoading,
```
iamoperand, line 430: nice. I think we can remove [this](https://github.com/opynfinance/squeeth-monorepo/pull/886/files#diff-683cf1dadaa899aca98cf44d3316f0ecf254c9461253cf711576b771b532fd73R356) condition now

iamoperand (approved):
tested out the functionality on Goerli and it is working as expected 🥳

just left a few minor comments, lmk what you think!

packages/frontend/src/state/crab/hooks.ts
rushikeshakhare: Cant find the condition you are pointing at

alexisgauba:
Hmm, so looks like it eventually pulls the right position, but tbh it looks like its glitching for a while before it loads the position, which made feel a bit worried about what was going on as a user -> maybe we want a loader icon with text like "updating position" and have it just be continuous loading vs. flashing in and out until the proper amount is added


https://user-images.githubusercontent.com/13340486/210490948-da39c3df-66cc-4b1a-80f8-b0307a403ee1.mov

alexisgauba (approved):

PR #908, opened 2023-01-09
 Exclude fee from PI calculation

KMKoushik:
# Task:

## Description

Fixes ENG-1294

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Testing code


aleone (approved):
generally lgtm - think we should probably make it less magic numbery and mabye confirm the calc is 100% correct (it might be that we are multiplying the 100.3% at wrong step), but its a small difference and ok for a fix for me.

rushikeshakhare (approved):

PR #919, opened 2023-01-16
Fix: Add eth Amount returned in bull deposited amount calculation

KMKoushik:
Fixes ENG-1293

## Type of change

- [ ] New feature
- [X] Bug fix
- [ ] Testing code

### DIFF
--- a/packages/subgraph/subgraph.template.yaml
+++ b/packages/subgraph/subgraph.template.yaml
@@ -1,4 +1,4 @@
-specVersion: 0.0.4
+specVersion: 0.0.6
 schema:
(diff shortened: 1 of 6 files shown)


packages/subgraph/subgraph.template.yaml
```diff
@@ -1,4 +1,4 @@
-specVersion: 0.0.4
+specVersion: 0.0.6
```
KMKoushik, line 1: To get all the logs of a transaction. More details [here](https://github.com/graphprotocol/graph-node/pull/3373)

alexisgauba (approved):

PR #927, opened 2023-01-17
Feat: queue WETH

haythemsellami:
# Task: queue WETH

**Should be reviewed/merged after #925**

## Description

This PR add the functionality of queuing WETH token.

Fixes ENG-1368

## Type of change

- [X] New feature
- [ ] Bug fix
- [ ] Testing code


packages/zen-bull-netting/src/ZenBullNetting.sol
```diff
@@ -35,6 +37,17 @@ contract ZenBullNetting is Ownable, EIP712 {
     /// @dev twap period to use for auction calculations
     uint32 public auctionTwapPeriod;
 
+    /// @dev WETH token address
+    address private weth;
+
+    /// @dev array of deposit receipts
+    Receipt[] public deposits;
+
+    /// @dev WEth amount to deposit for an address
```
alpinechicken, line 46: *WETH

packages/zen-bull-netting/test/unit-test/ZenBullNettingUnit.t.sol
```diff
@@ -49,16 +58,14 @@ contract ZenBullNettingUnit is ZenBullNettingBaseSetup {
     }
 
     function testSetMinWethAmount() public {
-        vm.prank(owner);
-        zenBullNetting.setMinWethAmount(5e18);
-
-        assertEq(zenBullNetting.minWethAmount(), 5e18);
+        _setMinWethAmount();
+        assertEq(zenBullNetting.minWethAmount(), minWeth);
```
alpinechicken, line 62: Curious why setting this as state but not minBullAmount etc

alpinechicken (approved):
lgtm

packages/zen-bull-netting/src/ZenBullNetting.sol
```diff
@@ -49,17 +62,35 @@ contract ZenBullNetting is Ownable, EIP712 {
         bytes32 s;
     }
 
+    /// @dev receipt used to store deposits and withdraws
+    struct Receipt {
+        /// @dev address of the depositor or withdrawer
+        address sender;
+        /// @dev usdc amount to queue for deposit or crab amount to queue for withdrawal
```
aleone, line 69: adjust comment

packages/zen-bull-netting/src/ZenBullNetting.sol
```diff
@@ -140,4 +171,35 @@ contract ZenBullNetting is Ownable, EIP712 {
 
         otcPriceTolerance = _otcPriceTolerance;
     }
+
+    /**
+     * @notice queue USDC for deposit into crab strategy
+     * @param _amount USDC amount to deposit
+     */
+    function queueWeth(uint256 _amount) external {
```
aleone, line 179: I think we probably want users to queue ETH generally (with maybe the advanced option to queue ETH)

I think could create a wrapper function that just wraps and then calls queueWeth.

Right now deposits and withdrawals are in ETH not WETH for zen, so should be consistent.

If you want can be in another PR, but lets add a linear issue/ticket for it and add to spec.

packages/zen-bull-netting/test/unit-test/ZenBullNettingUnit.t.sol
aleone: @alpinechicken why you mean? just the decision of why to create a function/have a variable have minWethAmount?

packages/zen-bull-netting/test/unit-test/ZenBullNettingUnit.t.sol
```diff
@@ -139,4 +146,39 @@ contract ZenBullNettingUnit is ZenBullNettingBaseSetup {
...
+        assertEq(
+            IERC20(WETH).balanceOf(address(zenBullNetting)) - zenBullNettingWethBalanceBefore,
+            amount
+        );
+        assertEq(IERC20(WETH).balanceOf(user1) + amount, user1WethBalanceBefore);
+        assertEq(zenBullNetting.depositsQueued(), amount);
+    }
+
+    function testQueueWethWhenAmountLessThanMinAmount() public {
+        _setMinWethAmount();
```
aleone, line 170: why do you do here, but not in ``testQueueWeth()`` ?

packages/zen-bull-netting/test/unit-test/ZenBullNettingUnit.t.sol
```diff
@@ -139,4 +146,39 @@ contract ZenBullNettingUnit is ZenBullNettingBaseSetup {
...
+        vm.stopPrank();
+
+        assertEq(zenBullNetting.wethBalance(user1), amount);
+        assertEq(
+            IERC20(WETH).balanceOf(address(zenBullNetting)) - zenBullNettingWethBalanceBefore,
+            amount
+        );
+        assertEq(IERC20(WETH).balanceOf(user1) + amount, user1WethBalanceBefore);
+        assertEq(zenBullNetting.depositsQueued(), amount);
+    }
```
aleone, line 167: should also check if the receipt is done correctly? 

ie msg.sender, timestamp recorded etc?

aleone (approved):
Generally lgtm, think we need to have ETH deposits enabled, but can be in a PR.

Think we should ideally test the receipt is recorded correctly (is not just amt but timestamp, msg.sender).

packages/zen-bull-netting/test/unit-test/ZenBullNettingUnit.t.sol
haythemsellami: Ah u mean `minWeth` variable, just because I needed to add before some tests to make sure the min amount is set, will add also `minBull` later in the Bull queueing PR, and maybe later move all of this inside the setup function...

packages/zen-bull-netting/test/unit-test/ZenBullNettingUnit.t.sol
haythemsellami: Yeah didn't add those specific asserts as the code is the same in Crab netting, can add them later

packages/zen-bull-netting/test/unit-test/ZenBullNettingUnit.t.sol
haythemsellami: Here testing the revert case

PR #929, opened 2023-01-17
Feat: dequeue WETH

haythemsellami:
# Task: dequeue WETH

**Should be reviewed/merged after #927**

## Description

Fixes ENG-1369

## Type of change

- [X] New feature
- [ ] Bug fix
- [ ] Testing code


alpinechicken (approved):
lgtm

aleone (approved):
lgtm, but think we should add the comments for the error strings as we go through these PRs vs at the end so can make sure that the right error string matches what code is reverting on

PR #941, opened 2023-01-20
Feat: withdraw auction

haythemsellami:
# Task: withdraw auction

## Description

Fixes ENG-1366

## Type of change

- [X] New feature
- [ ] Bug fix
- [ ] Testing code


sanandnarayan (approved):
zenbullnetting.sol looks good to me

sanandnarayan (approved):

packages/zen-bull-netting/src/ZenBullNetting.sol
```diff
@@ -427,6 +484,143 @@ contract ZenBullNetting is Ownable, EIP712 {
...
+                    _params.orders[i].trader, address(this), _params.orders[i].quantity
+                );
+            } else {
+                IERC20(oSqth).transferFrom(_params.orders[i].trader, address(this), toExchange);
+                break;
+            }
+        }
+
+        uint256 usdcToRepay = _params.withdrawsToProcess
+            * IEulerSimpleLens(eulerLens).getDTokenBalance(usdc, zenBull) / bullTotalSupply;
```
aleone, line 524: Think this has different order of operations again (and not sure if any differences to use lens), so it might have a difference due to precision? 

contract calculates using this ` _bullShare.wmul(IEulerDToken(dToken).balanceOf(address(this)));`

packages/zen-bull-netting/src/ZenBullNetting.sol
```diff
@@ -427,6 +484,143 @@ contract ZenBullNetting is Ownable, EIP712 {
...
+            }
+            if (withdraw.amount <= remainingWithdraws) {
+                // full usage
+                remainingWithdraws -= withdraw.amount;
+                zenBullBalance[withdraw.sender] -= withdraw.amount;
+
+                // send proportional usdc
+                ethAmount = withdraw.amount * ethToWithdrawers / _params.withdrawsToProcess;
+
+                payable(withdraw.sender).sendValue(ethAmount);
```
aleone, line 583: has reentrancy, will need to explore this. we fwd all gas, maybe we don't want to use sendValue?

packages/zen-bull-netting/src/ZenBullNetting.sol
```diff
@@ -427,6 +484,143 @@ contract ZenBullNetting is Ownable, EIP712 {
...
+                j++;
+                continue;
+            }
+            if (withdraw.amount <= remainingWithdraws) {
+                // full usage
+                remainingWithdraws -= withdraw.amount;
+                zenBullBalance[withdraw.sender] -= withdraw.amount;
+
+                // send proportional usdc
+                ethAmount = withdraw.amount * ethToWithdrawers / _params.withdrawsToProcess;
```
aleone, line 581: I think this is ok as it will round down, but its possible that folks get less ETH due to precision, just noting

example, 2 wei * 100 wei / 300 wei rounds down to 0

its ok as long as it doesn't round up

aleone (approved):
solidity code left a few comments - need to review tests tmrw - can merge once have a look at comments and I'll review tests after merge

packages/zen-bull-netting/src/ZenBullNetting.sol
haythemsellami: What do u mean "has reentrancy" ?

PR #947, opened 2023-01-24
Pnl bands

KMKoushik:
# Task:

## Description

Fixes (linear-task)

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Testing code


alexisgauba (changes requested):
Awesome! Some comments: 

- we should also show deposit selector when wallet connected show user knows what date the chart is referring to and also has the ability to play around even when connected
- we should align deposit date with top of graph 
Screen Shot 2023-01-25 at 4 41 02 PM

- there should be more space between deposit date & next hedge in 
Screen Shot 2023-01-25 at 4 40 33 PM

- realize it looks a bit weird that they are diff fonts, maybe same font size of deposit date and next hedge in, could make next hedge in smaller (same size as the deposit date numbers) and that could help with the alignment as well 

- Looks like graph not updating for dates before the hedge vs. after the hedge
Screen Shot 2023-01-25 at 4 44 43 PM
Screen Shot 2023-01-25 at 4 44 36 PM

alexisgauba (changes requested):
- Looks like a weird UI thing here 
Screen Shot 2023-01-26 at 11 02 20 AM

- Looks like a loading thing where for a historical date the bands get wider and then thinner -  initially it was actually only giving me those wider bands but eventually it started loading the thinner ones 

https://user-images.githubusercontent.com/13340486/214926977-3e5ff55b-c0f9-47fb-946d-6621be4407c0.mov



- Prob worth checking with @aleone to make sure calcs all good

alexisgauba:
We should do this for bull as well! Just noticed its not done there yet - can be finalized for crab first and then copy / pasted over, but imo we shld launch with both

alexisgauba:
Also for users already in the strategy, the deposit date should be the date they deposited when their wallet is connected. This account has a test position but its showing today's date not deposit date 

Screen Shot 2023-01-27 at 5 14 44 PM

packages/frontend/src/utils/strategyPayoff.ts
```diff
@@ -0,0 +1,188 @@
...
+  collat = collat + bullExcessEth
+
+  const portfolioValueInETH = collat - shortAmt * oSqthPrice
+  const portfolioValueInUSD = portfolioValueInETH * ethPriceAtHedge
+
+  const starting = new BigNumber(-percentRange)
+  const increment = new BigNumber(0.05)
+  const ending = new BigNumber(percentRange)
+
+  const ethRate = Math.exp(ethSupplyApy / (365 / time))
```
alpinechicken, line 137: Calling this rate is a bit confusing, maybe ethFutureValue to be specific

alpinechicken: same with usdRate

packages/frontend/src/utils/strategyPayoff.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  return {
+    dataPoints,
+    lowerPriceBandForProfitability: getData(ethPriceAtHedge + thresholdLower),
+    upperPriceBandForProfitability: getData(ethPriceAtHedge + thresholdUpper),
+    currentProfit: getData(currentEthPrice),
+  }
+}
+
+export const getBullProfitDataPoints = (
```
alpinechicken, line 104: maybe we should have some naming so it's clear that this is bull *excess* profit rather than bull total profit. getBullExcessProfitDataPoints() ?

alpinechicken (approved):
bounds lgtm. Couple of suggestions for naming to make a bit clearer.

alexisgauba:
Looks like the tooltip is still doin dis weird thing, maybe we want the tooltip above the background stuffs? 

Screen Shot 2023-02-02 at 9 28 15 PM

PR #960, opened 2023-01-31
fix: landing page spacing issues

rushikeshakhare:
# Task:

Fix Landing page spacing

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Testing code


alexisgauba (changes requested):
think we can prob also reduce some space between the numbers and the "strategies" header! seems like with this change there's less space between the hero and stats but more between the stats and strategies

iamoperand (approved):

KMKoushik (approved):

PR #956, opened 2023-01-30
Chore: deployment scripts

haythemsellami:
# Task: deployment scripts

**Should be reviewed/merged after #954**

## Description

Instructions how to deploy in the README.

Fixes ENG-1373

## Type of change

- [X] New feature
- [ ] Bug fix
- [ ] Testing code
- [ ] Document update or config files

PR #964, opened 2023-02-02
Fix: mul and div orders in withdraw auction

haythemsellami:
# Task: mul and div orders in withdraw auction

**Should be reviewed/merged after #956**

## Description

Fixes ENG-1439

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Testing code

### DIFF
--- a/packages/zen-bull-netting/src/NettingLib.sol
+++ b/packages/zen-bull-netting/src/NettingLib.sol
@@ -172,9 +172,20 @@ library NettingLib {
         returns (uint256)
     {
         uint256 bullTotalSupply = IERC20(zenBull).totalSupply();
-        uint256 crabAmount =
-            withdrawsToProcess * IZenBullStrategy(zenBull).getCrabBalance() / bullTotalSupply;
         (, uint256 crabDebt) = IZenBullStrategy(zenBull).getCrabVaultDetails();
-        return (crabAmount * crabDebt / IERC20(crab).totalSupply());
+        uint256 share = div(withdrawsToProcess, bullTotalSupply);
+        uint256 crabAmount = mul(share, IZenBullStrategy(zenBull).getCrabBalance());
+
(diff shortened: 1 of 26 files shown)


packages/zen-bull-netting/src/NettingLib.sol
```diff
@@ -172,9 +172,20 @@ library NettingLib {
...
-            withdrawsToProcess * IZenBullStrategy(zenBull).getCrabBalance() / bullTotalSupply;
         (, uint256 crabDebt) = IZenBullStrategy(zenBull).getCrabVaultDetails();
-        return (crabAmount * crabDebt / IERC20(crab).totalSupply());
+        uint256 share = div(withdrawsToProcess, bullTotalSupply);
+        uint256 crabAmount = mul(share, IZenBullStrategy(zenBull).getCrabBalance());
+
+        return div(mul(crabAmount, crabDebt), IERC20(crab).totalSupply());
+    }
+
+    function mul(uint256 x, uint256 y) internal pure returns (uint256) {
```
alpinechicken, line 182: why not just use wmul and wdiv?

alpinechicken (approved):
lgtm

packages/zen-bull-netting/src/NettingLib.sol
aleone: @alpinechicken seems like same implementation if you look at strategy math vs above, can we just not use safe math but have the order of operations just be the same and get the same result?  it seems using the strategy math implementation does waste some gas due to overflow prevention, right?

aleone (approved):

packages/zen-bull-netting/src/NettingLib.sol
haythemsellami: yeah also the StrategyMath is written with a diff solidity version, if I just import it, it will not work

PR #966, opened 2023-02-02
feat: amplitude proxy

rushikeshakhare:
# Task:
Fixes ENG-1388

Remove cookie consent

Legal modal update

Screenshot 2023-02-02 at 9 52 11 PM

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Testing code


iamoperand (approved):

tamtam3000:
1.  Please confirm that we are still using Amplitude, just not the Amplitude cookie. 

If it's correct that the FE is still using Amplitude (but not the cookie), change to: "The app logs anonymized usage statistics to improve over time" 

2. In Sub- title, change to: This app uses the following third-party APIs , Not "The" . In subsequent descriptions for the APIs start with "The", instead or rotating between this and thus throughout. 

3. Also change Sentry description - there is a nit, and unnecessary "the" before Sentry. 

"The app logs blocked address events using Sentry"

4. Remove the period (.) in the Amplitude description and the end of the sentence. 

5. Last: The UI is funky because I cannot scroll down to look at the Graph description, it just scrolls the background to the pop-up, which means the user cannot see the full description of APIs and cannot click on the learn more button at the bottom of the pop-up. The pop-up itself, not the background, should have a scroll function.

PR #965, opened 2023-02-02
Feat: add bot address to automatically net deposits with withdraws 

haythemsellami:
# Task: add bot address to automatically net deposits with withdraws 

**Should be reviewed.merged after #964**
 
## Description

Fixes ENG-1441

## Type of change

- [X] New feature
- [ ] Bug fix
- [ ] Testing code
- [ ] Document update or config files

PR #968, opened 2023-02-02
Feat: cancel nonce

haythemsellami:
# Task: cancel nonce

**Should be reviewed/merged after #965**

## Description

Fixes ENG-1440

## Type of change

- [X] New feature
- [ ] Bug fix
- [ ] Testing code


alpinechicken (approved):
lgtm

packages/zen-bull-netting/test/unit-test/CancelNonce.t.sol
```diff
@@ -0,0 +1,33 @@
...
+    function setUp() public override {
+        ZenBullNettingBaseSetup.setUp();
+
+        traderPk = 0xA12CD;
+        trader1 = vm.addr(traderPk);
+
+        vm.label(trader1, "trader1");
+    }
+
+    function testCancelNonce() public {
```
aleone, line 25: Maybe better to test in the context of an order - ie flip a nonce to false and the auction fill should fail, but maybe that is tested already.

aleone (approved):

PR #990, opened 2023-03-16
Add Emergency withdraw contract

KMKoushik:


packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * @notice withdraw ETH deposited into crab
+     * @dev this will give the sender ZBEPR token as ownership for the ETH deposited in Euler pool
+     * @param _zenBullAmount ZenBull amount to redeem
+     * @param _maxEthForWPowerPerp max ETH to pay for flashswapped oSQTH amount
+     */
+    function withdraw(uint256 _zenBullAmount, uint256 _maxEthForWPowerPerp) external {
+        uint256 crabToRedeem =
+            _zenBullAmount.wdiv(_zenBullSupply).wmul(IZenBullStrategy(zenBull).getCrabBalance());
```
aleone, line 82: maybe reverse order of operations

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,130 @@
...
+            0,
+            abi.encodePacked(crabToRedeem, wPowerPerpToRedeem)
+        );
+
+        uint256 payout = IWETH9(weth).balanceOf(address(this));
+        IWETH9(weth).withdraw(payout);
+        payable(msg.sender).sendValue(payout);
+
+        emit Withdraw(
+            msg.sender, _zenBullAmount, crabToRedeem, wPowerPerpToRedeem, payout, _zenBullAmount
```
aleone, line 107: double _zenBullAmount emitted event

aleone (commented):
Generally lgtm, few small comments.  I want to improve testing though.  Some things to add:

- test to ensure that things that we want to fail will fail currently:
  - set strategy cap to 0
  - users can not deposit/withdraw in normal ZB
- test expected behavior if euler unbricks - we can test this by running this fork on a pre-exploit block and withdrawing some users, but not everyone via emergency contract
   - test that if strategyCap = 0, users can not deposit() in normal ZB
   - test that normal ZB withdraw() after others have done emergency shutdown withdraw() will give you less ETH than you should get (this is expected behavior - no one should use ZB withdraw() after this ES contract has been deployed and used, even if euler unpauses.

Maybe others, anyone have other tests we should add?

packages/zen-bull-vault/src/EmergencyWithdraw.sol
haythemsellami: One for `zenBullAmountRedeemed` and the other `eulerRecoveryTokenAmount`

packages/zen-bull-vault/test/integration-test/EmergencyWithdraw/Withdraw.t.sol
```diff
@@ -0,0 +1,116 @@
...
+        uint256 user1BullBalanceAfter = IERC20(ZEN_BULL).balanceOf(user1);
+        uint256 user1EthBalanceAfter = address(user1).balance;
+        uint256 user1RecoveryTokenBalanceAfter = IERC20(emergencyWithdraw).balanceOf(user1);
+
+        assertEq(bullSupplyAfter, bullSupplyBefore - user1BullBalanceBefore);
+        assertEq(user1BullBalanceAfter, 0);
+        assertEq(
+            user1RecoveryTokenBalanceAfter - user1RecoveryTokenBalanceBefore, user1BullBalanceBefore
+        );
+        assertGt(user1EthBalanceAfter, user1RecoveryTokenBalanceBefore);
```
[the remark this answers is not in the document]
haythemsellami: removed

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,130 @@
...
+        require(msg.sender == weth, "Can't receive ETH from this sender");
+    }
+
+    /**
+     * @notice withdraw ETH deposited into crab
+     * @dev this will give the sender ZBEPR token as ownership for the ETH deposited in Euler pool
+     * @param _zenBullAmount ZenBull amount to redeem
+     * @param _maxEthForWPowerPerp max ETH to pay for flashswapped oSQTH amount
+     */
+    function withdraw(uint256 _zenBullAmount, uint256 _maxEthForWPowerPerp) external {
```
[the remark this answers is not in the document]
haythemsellami: renamed to `emergencyWithdrawEthFromCrab`

packages/zen-bull-vault/src/EmergencyWithdraw.sol
haythemsellami: I followed the order in the ZenBull implementation

packages/zen-bull-vault/src/EmergencyWithdraw.sol
aleone: ah I see

packages/zen-bull-vault/test/integration-test/EmergencyWithdraw/EmergencyWithdrawEthFromCrab.t.sol
```diff
@@ -0,0 +1,202 @@
...
+        assertEq(bullSupplyAfter, bullSupplyBefore - user1BullBalanceBefore);
+        assertEq(user1BullBalanceAfter, 0);
+        assertEq(
+            user1RecoveryTokenBalanceAfter - user1RecoveryTokenBalanceBefore, user1BullBalanceBefore
+        );
+        assertEq(
+            ethToWithdrawFromCrab.sub(maxWethForOsqth),
+            user1EthBalanceAfter.sub(user1EthBalanceBefore)
+        );
+    }
```
alpinechicken, line 114: Possibly also check that bull crab balance changes as expected (though it seems hard for it not to).

alpinechicken (approved):
lgtm - one suggestion for test but probably not necessary

haythemsellami:
> Generally lgtm, few small comments. I want to improve testing though. Some things to add:
> 
> * test to ensure that things that we want to fail will fail currently:
>   
>   * set strategy cap to 0
>   * users can not deposit/withdraw in normal ZB
> * test expected behavior if euler unbricks - we can test this by running this fork on a pre-exploit block and withdrawing some users, but not everyone via emergency contract
>   
>   * test that if strategyCap = 0, users can not deposit() in normal ZB
>   * test that normal ZB withdraw() after others have done emergency shutdown withdraw() will give you less ETH than you should get (this is expected behavior - no one should use ZB withdraw() after this ES contract has been deployed and used, even if euler unpauses.
> 
> Maybe others, anyone have other tests we should add?

Added those tests other than the deposit() when cap is 0, as this is already tested in Deposit tests

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,131 @@
...
+        require(msg.sender == weth, "Can't receive ETH from this sender");
+    }
+
+    /**
+     * @notice withdraw ETH deposited into crab
+     * @dev this will give the sender ZBEPR token as ownership for the ETH deposited in Euler pool
+     * @param _zenBullAmount ZenBull amount to redeem
+     * @param _maxEthForWPowerPerp max ETH to pay for flashswapped oSQTH amount
+     */
+    function emergencyWithdrawEthFromCrab(uint256 _zenBullAmount, uint256 _maxEthForWPowerPerp)
```
alpinechicken, line 79: A thought: It might be nice to have an alternative withdraw where they provide oSQTH rather than needing to trade through the pool. 

They could achieve the sam thing by putting their oSQTH in a 1-tick LP and calling this, but more complicated and worse gas.

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -109,6 +152,84 @@ contract EmergencyWithdraw is ERC20, UniFlash {
...
+        uint24 _poolFee
+    ) external {
+        uint256 recoverShare = _recoveryTokenAmount.wdiv(totalSupply());
+        uint256 usdcToRepay = recoverShare.wmul(IEulerDToken(dToken).balanceOf(zenBull));
+        uint256 wethToWithdraw =
+            recoverShare.wmul(IEulerEToken(eToken).balanceOfUnderlying(zenBull));
+
+        require(
+            wethToWithdraw <= MAX_WETH_PER_DEBT_REPAY,
+            "WETH to withdraw is greater than max per repay"
```
alpinechicken, line 175: What are we trying to prevent here?

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -109,6 +152,84 @@ contract EmergencyWithdraw is ERC20, UniFlash {
...
+        );
+
+        uint256 ethUsdPrice = UniOracle._getTwap(ethUSDCPool, weth, usdc, 420, false);
+
+        require(
+            _maxEthForUsdc <= ethUsdPrice.wmul((ONE.add(LIMIT_PRICE_TOLERANCE))),
+            "Max ETH for USDC greater than max price tolerance"
```
alpinechicken, line 182: Just for safety so they don't accidentally get rekt buying USDC?

alpinechicken: ah I see it's so everyone doesn't get rekt

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,268 @@
...
+        );
+
+        ethDividend[msg.sender] = ethDividend[msg.sender].add(_recoveryTokenAmount);
+        totalDividend = totalDividend.add(_recoveryTokenAmount);
+        _burn(msg.sender, _recoveryTokenAmount);
```
aleone, line 187: this should be done in ```withdrawEth()``

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,268 @@
...
+            usdcToRepay,
+            usdcToRepay.wmul(_maxEthForUsdc.mul(1e12)),
+            uint8(FLASH_SOURCE.EMERGENCY_REPAY_EULER_DEBT),
+            abi.encodePacked(usdcToRepay, wethToWithdraw)
+        );
+
+        IWETH9(weth).withdraw(IERC20(weth).balanceOf(address(this)));
+
+        if (
+            (IEulerEToken(eToken).balanceOfUnderlying(zenBull) == 0)
```
aleone, line 202: is it possible that interest is accrued and we have like 2 wei left or something in contract

we only touch the protocol when we repay/withdraw, is it possible that balanceOf and balanceOfUnderlying() changes between initial call and after the repay

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,253 @@
...
+        uint256 _wethToWithdraw,
+        uint256 _maxEthForUsdc,
+        uint24 _poolFee
+    ) external {
+        require(
+            _wethToWithdraw <= MAX_WETH_PER_DEBT_REPAY,
+            "WETH to withdraw is greater than max per repay"
+        );
+
+        uint256 ratio = _wethToWithdraw.wdiv(IEulerEToken(eToken).balanceOfUnderlying(zenBull));
```
aleone, line 174: might be impossible to make this ratio 100%, because the eToken balanceOfUnderlying may change constantly (even just in the time from off chain calc to onchain execution)

I know I said fine to do either via explicit amounts, but maybe its better to have the input be a percentage, so can start with 10%, then 20%, then 50%, then 100% and explicitly try to withdraw 100%.

aleone: And you can calculate the amounts in the same way, as usdcToRepay (and for wethToWithdraw), using the explicit ratio.

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,253 @@
...
+            _maxEthForUsdc <= ethUsdPrice.wmul((ONE.add(LIMIT_PRICE_TOLERANCE))),
+            "Max ETH for USDC greater than max price tolerance"
+        );
+
+        _exactOutFlashSwap(
+            weth,
+            usdc,
+            _poolFee,
+            usdcToRepay,
+            usdcToRepay.wmul(_maxEthForUsdc.mul(1e12)),
```
aleone, line 188: maybe make magic # more explicit

aleone: this should be usdcToRepay.mul(1e12).wdiv(_maxEthForUsdc), right?  its an amount of ETH in max

aleone: ie 1000e6 (1k$) assuming 1k$ eth price -> 1e6*1e12/1000e18 = 1eth max in

packages/zen-bull-vault/src/EmergencyWithdraw.sol
aleone: still concerned about this, can we look at euler and see if this is a problem and also try to test maybe using past blocks or something?

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,253 @@
...
+            uint8(FLASH_SOURCE.EMERGENCY_REPAY_EULER_DEBT),
+            abi.encodePacked(usdcToRepay, _wethToWithdraw)
+        );
+
+        IWETH9(weth).withdraw(IERC20(weth).balanceOf(address(this)));
```
aleone, line 193: good to withdraw this every time as payout above is determined by balanceOf, so as long as we withdraw everytime, its fine

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,253 @@
...
+        emit EmergencyRepayEulerDebt(
+            msg.sender, usdcToRepay, _wethToWithdraw, _maxEthForUsdc, ethWithdrawalActivated
+        );
+    }
+
+    /**
+     * @notice withdraw ETH from this contract, will revert if ZenBull still have some debt not repaid
+     */
+    function withdrawEth(uint256 _recoveryTokenAmount) external {
+        require(ethWithdrawalActivated);
```
aleone, line 211: maybe have an error message here

aleone (changes requested):
main concerns are hitting 100% ratio and accrued debt/interest in euler
some other small comments

want to test what happens if we unwind euler with only X% of users having withdrawn crab and ensure they get the correct crab ETH value out when withdrawing as well as their correct euler share (as this is 100% the case that will happen)

packages/zen-bull-vault/src/EmergencyWithdraw.sol
haythemsellami: But maxEthForUsd is how much ETH to pay for 1 USDC right? at least that's how I did it

packages/zen-bull-vault/test/e2e/EmergencyWithdrawScenario.t.sol
```diff
@@ -0,0 +1,268 @@
...
+            IERC20(WPOWERPERP).approve(ZEN_BULL, wPowerPerpToRedeem);
+            ZenBullStrategy(ZEN_BULL).withdraw(bullAmountToWithdraw);
+            vm.stopPrank();
+
+            uint256 user2EthBalanceAfter = address(user2).balance;
+            normalWithdrawPayoutAfterEmergencyWithdraws =
+                user2EthBalanceAfter.sub(user2EthBalanceBefore);
+        }
+
+        assertLt(
```
aleone, line 213: Ideally for this I'd like to check that a user withdrawing via emergency withdraw contract after the user mistakenly withdraws from with normal withdraw() function has no issues and gets a bit more eth than is expected.  A user calling withdraw() after this contract is used is effectively donating some of their funds to everyone else in the strategy, so it should never be used.

aleone (commented):
some more test comments

an additional test I think we should add, is in e2e - test that a user pre euler exploit will receive the same payout if calling emergencyWithdrawEthFromCrab() and then emergencyWithdrawEthFromCrabTest() vs just calling withdraw().  Total ETH received by user should be the same either way.

packages/zen-bull-vault/src/EmergencyWithdraw.sol
aleone: ```UniOracle._getTwap(ethUSDCPool, weth, usdc, 420, false);``` returns USDC per 1 ETH scaled to 1e18

```        require(
            _maxEthForUsdc <= ethUsdPrice.wmul((ONE.add(LIMIT_PRICE_TOLERANCE))),
            "Max ETH for USDC greater than max price tolerance" ```
            
          
This also takes USDC per ETH prices.

**Review by aleone — CHANGES_REQUESTED** [review]

File: packages/zen-bull-vault/src/EmergencyWithdraw.sol (hunk 12 above)
Comment (haythemsellami) [reply packages/zen-bull-vault/src/EmergencyWithdraw.sol]: I updated it to take USDC price per 1 ETH

File: packages/zen-bull-vault/src/EmergencyWithdraw.sol [hunk 16]
```diff
@@ -0,0 +1,255 @@
...
+            weth,
+            usdc,
+            _poolFee,
+            usdcToRepay,
+            usdcToRepay.mul(WETH_DECIMALS_DIFF).wmul(_maxEthForUsdc),
+            uint8(FLASH_SOURCE.EMERGENCY_REPAY_EULER_DEBT),
+            abi.encodePacked(usdcToRepay, wethToWithdraw)
+        );
+
+        IWETH9(weth).withdraw(IERC20(weth).balanceOf(address(this)));
```
alpinechicken, line 195: When we accumulate eth here can't it be fully withdrawn by anyone calling emergencyWithdrawEthFromCrab?

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,255 @@
...
+            ethWithdrawalActivated = true;
+        }
+
+        emit EmergencyRepayEulerDebt(
+            msg.sender, _ratio, usdcToRepay, wethToWithdraw, _maxEthForUsdc, ethWithdrawalActivated
+        );
+    }
+
+    /**
+     * @notice withdraw ETH from this contract, will revert if ZenBull still have some debt not repaid
```
alpinechicken, line 210: *still has some debt not repaid

aleone (changes requested):
some addtl cmments

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,255 @@
...
+    {
+        uint256 usdcToRepay = _ratio.wmul(IEulerDToken(dToken).balanceOf(zenBull));
+        uint256 wethToWithdraw = _ratio.wmul(IEulerEToken(eToken).balanceOfUnderlying(zenBull));
+
+        require(
+            wethToWithdraw <= MAX_WETH_PER_DEBT_REPAY,
+            "WETH to withdraw is greater than max per repay"
+        );
+
+        uint256 ethUsdPrice = UniOracle._getTwap(ethUSDCPool, usdc, weth, 420, false);
```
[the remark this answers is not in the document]
alpinechicken: It confused me initially. Def easier to check with usdc per eth.

sanandnarayan (changes requested):

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,255 @@
...
+        crab = _crab;
+        zenBull = _zenBull;
+        weth = _weth;
+        usdc = _usdc;
+        wPowerPerp = _wPowerPerp;
+        ethUSDCPool = _ethUSDCPool;
+        eToken = _eToken;
+        dToken = _dToken;
+
+        zenBullTotalSupplyForCrabWithdrawal = IERC20(_zenBull).totalSupply();
```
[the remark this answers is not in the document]
aleone: @sanandnarayan @haythemsellami we need to do the same logic for the ```zenBullTotalSupplyForEulerWithdrawal``` variable, where we adjust it by a ```redeemedZenBullEulerRecoveryAmount``` and current ```totalSupply()```

If a user calls withdraw() (after the above fixes):
Initial state -> Crab in ZB = 100, Eth in Euler = 200, Usdc in Euler = 100, ETH px = 1, ZB totalSupply() = 200, redeemedZenBullAmount = 0, zenBullTotalSupplyForEulerWithdrawal = 200

Step 1: 100 ZB emergency withdraw -> (100/(200-0)) * 100 -> receive 50 crab
State after step 1: 
Crab in ZB = 50, Eth in Euler = 200, Usdc in Euler = 100, ETH px = 1, ZB totalSupply() = 200, redeemedZenBullAmount = 100, zenBullTotalSupplyForEulerWithdrawal = 200

Step 2 (euler works again): 50 ZB withdraw() -> (50/200)*50 -> receive 12.5 crab (50/200), (50/200) * 200 ETH - (50/200) * 100 USDC -> repay 25 USDC, receive 50 ETH, at ETH=1, receive 25 ETH. This user should have ideally received 25 crab and 25 eth, but only received 12.5 crab and 25 eth.
State after step 2:
Crab in ZB = 37.5, Eth in Euler = 150, Usdc in Euler = 75, ETH px = 1, ZB totalSupply() = 150, redeemedZenBullAmount = 100, zenBullTotalSupplyForEulerWithdrawal = 200

Step 3: 50 ZB emergency withdrawn -> 37.5 * 50 / (150-100) -> receive ETH from 37.5 crabs. Should have received only 25, but collects the extra donated by user in step 2.
State after step 3: 
Crab in ZB = 0, Eth in Euler = 150, Usdc in Euler = 75, ETH px = 1, ZB totalSupply() = 150, redeemedZenBullAmount = 150, zenBullTotalSupplyForEulerWithdrawal = 200

Step 4: We unwind euler trade.  
State after step 4 that changes: Eth in contract = 75

Step 5: User with 100 ZBRecoveryTokens tries to withdraw their ETH. They should receive 50 ETH as payout (as they were 50% of strategy when had 200 ETH of collateral and 100 USDC of debt).  (100 / 200) * 75 = 37.5.  So they (and every following withdrawer gets less than they should).

Fix is same here.  We need to have this user claim 2/3 or 100/150 of the ETH in the strategy, and can do in the same way by refreshing the totalSupply() each time we calc the share and keep track of how much zb recovery tokens hve been redeemed.

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,262 @@
...
+    /// @dev eth/usdc uniswap pool address
+    address internal immutable ethUSDCPool;
+    address internal immutable eToken;
+    address internal immutable dToken;
+
+    /// @dev total amount of redeemed ZenBull through emergencyWithdrawEthFromCrab
+    uint256 public redeemedZenBullAmountForCrabWithdrawal;
+    /// @dev ZenBull total supply amount at the time of this contract deployment, used for Euler withdrawal calc
+    uint256 public redeemedRecoveryAmountForEulerWithdrawal;
+    /// @dev if true, enable ETH withdrawal from this contract
```
aleone, line 50: maybe more clearer to say something like enable ETH withdrawal after 100% of Euler debt and collateral has been removed

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,262 @@
...
+    ) ERC20("ZenBullEulerRecovery", "ZBER") UniFlash(_factory) {
+        crab = _crab;
+        zenBull = _zenBull;
+        weth = _weth;
+        usdc = _usdc;
+        wPowerPerp = _wPowerPerp;
+        ethUSDCPool = _ethUSDCPool;
+        eToken = _eToken;
+        dToken = _dToken;
+
```
aleone, line 103: lets delete this comment

packages/zen-bull-vault/src/EmergencyWithdraw.sol
```diff
@@ -0,0 +1,262 @@
...
+        require(
+            wethToWithdraw <= MAX_WETH_PER_DEBT_REPAY,
+            "WETH to withdraw is greater than max per repay"
+        );
+
+        uint256 ethUsdcPrice = UniOracle._getTwap(ethUSDCPool, weth, usdc, 420, false);
+
+        require(
+            _limitPriceUsdcPerEth <= ethUsdcPrice.wmul((ONE.add(LIMIT_PRICE_TOLERANCE))),
```
aleone, line 183: This isn't right

ethUsdcPrice is something like 1000 usdc per 1 eth
Limit price tolerance is something like 1%

We are selling eth to buy usdc, so we need to make sure the limit price is above

``` _limitPriceUsdcPerEth >= ethUsdcPrice.wmul((ONE.sub(LIMIT_PRICE_TOLERANCE)))```

minLimitPrice shouild be something like 1000 * (100%-1%) = 990
so if I used a limit price of 980 it should revert, but 990 should be allowed

**Review by aleone — CHANGES_REQUESTED** [review]
main comment is euler limit price isn't correct, small other comments.

still concerned with the euler accrued interested. tried messing with tests to make there be a lot of time between unwinds, but it seems to work, will continue to look at euler github and test more on this point.

**Review by haythemsellami — APPROVED** [review]

**Review by aleone — APPROVED** [review]

File: packages/zen-bull-vault/src/EmergencyWithdraw.sol [hunk 23]
```diff
@@ -24,7 +24,7 @@ contract EmergencyWithdraw is ERC20, UniFlash {
     /// @dev WETH decimals - USDC decimals
     uint256 internal constant WETH_DECIMALS_DIFF = 1e12;
     uint256 public constant MAX_WETH_PER_DEBT_REPAY = 100 ether;
-    uint256 public constant LIMIT_PRICE_TOLERANCE = 1e16; // 1%
+    uint256 public constant LIMIT_PRICE_TOLERANCE = 5e15; // 0.3%
```
alpinechicken: 0.5% in comment

packages/zen-bull-vault/test/integration-test/EmergencyWithdraw/EmergencyRepayEulerDebtTest.t.sol
```diff
@@ -73,18 +73,18 @@ contract EmergencyRepayEulerDebtTest is Test {
 
     function testEmergencyRepayEulerDebt() public {
         uint256 ethLimitPrice =
-            UniOracle._getTwap(ETH_USDC_POOL, WETH, USDC, 420, false).wmul((ONE.sub(1e16)));
+            UniOracle._getTwap(ETH_USDC_POOL, WETH, USDC, 420, false).wmul((ONE.sub(5e15)));
```
alpinechicken: 50bps still seems very high and probably sandwichable. I would think <10bps for any realistic size

sanandnarayan (approved):

PR #995, opened 2023-04-12
fix: fix crab migration withdraw

iamoperand:


KMKoushik (approved):

KMKoushik (changes requested):
Clicking max should only set migrated amount if available

PR #998, opened 2023-05-17
fix: short funding chart not showing recent data

nikkaroraa:
# Task:
funding not updating on f/e for short squeeth since apr 24th

## Description
Checkout the comments of the linear issue

Fixes ENG-1586

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Testing code

### DIFF
--- a/packages/frontend/src/components/Charts/ShortFundingChart.tsx
+++ b/packages/frontend/src/components/Charts/ShortFundingChart.tsx
@@ -3,9 +3,11 @@ import { Box, CircularProgress, createStyles, makeStyles, Tooltip } from '@mater
 import InfoIcon from '@material-ui/icons/InfoOutlined'
 import dynamic from 'next/dynamic'
 import React, { useState } from 'react'
+
 import CustomSwitch, { SwitchItem } from '@components/CustomSwitch'
-import { useNormHistory } from '@hooks/useNormHistory'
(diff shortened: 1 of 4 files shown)


KMKoushik (approved):

packages/frontend/src/components/Charts/ShortFundingChart.tsx
```diff
@@ -132,9 +141,18 @@ const FundingChart = () => {
       <div style={{ width: '100%' }}>
         <div className={classes.switchWrapper}>
           <CustomSwitch items={fundingTypes} value={fundingType} onChange={setFundingType} />
+
           {fundingType.id === 'funding' && (
             <CustomSwitch items={fundingDurations} value={fundingDuration} onChange={setFundingDuration} />
           )}
+
+          <Checkbox
```
KMKoushik, line 149: Can we change this to show full history?

aleone (changes requested):
Is there a way to have the default display include the most current data, even if we need to refetch to get the full history via checkbox?

Right now users don't see current data unless they check the box, but it seems the most important thing a user should care about is the most recent data.

Also the name of the checkbox, should maybe be "Show Full History" if thats what we can do.

nikkaroraa:
stale PR

PR #1009, opened 2023-10-31
if ipqs does not say vpn, then dont check again

sanandnarayan:
If we have too many requests to ipqs, then lets cache the positive responses also.

downside, the positive ips might later become VPNs and we might not catch them

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Testing code

### DIFF
--- a/packages/frontend/middleware.ts
+++ b/packages/frontend/middleware.ts
@@ -7,6 +7,9 @@ const redis = new Redis({
   token: process.env.UPSTASH_REDIS_REST_TOKEN!,
 })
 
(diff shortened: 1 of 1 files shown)


alpinechicken (approved):
lgtm - maybe remove logs. Also check overlaps with @nikaroraa prs.

aleone:
@nikkaroraa @alpinechicken should this be closed?

nikkaroraa:
closing this for now as i think its better of as it is now. but can reopen again if the need comes

PR #1012, opened 2023-11-28
feat: allow datacenter ips

nikkaroraa:
# Task:
Exclude datacenter IPs from VPNs

## Description
Fixes ENG-1944

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Testing code

### DIFF
--- a/packages/frontend/middleware.ts
+++ b/packages/frontend/middleware.ts
@@ -1,6 +1,8 @@
 import { NextRequest, NextResponse } from 'next/server'
 import { Redis } from '@upstash/redis'
+
 import { isVPN } from 'src/server/ipqs'
+import { BLOCKED_IP_VALUE } from 'src/constants'
 
 const redis = new Redis({
   url: process.env.UPSTASH_REDIS_REST_URL!,
@@ -26,7 +28,7 @@ export async function middleware(request: NextRequest) {
       console.error('Failed to get data from Redis:', error)
     }
 
-    const isIPBlocked = !!redisData
+    const isIPBlocked = redisData === BLOCKED_IP_VALUE
     console.log('ip', ip, isIPBlocked, url.protocol, url.host, '/blocked')
     if (isIPBlocked && url.pathname !== '/blocked') {
       return NextResponse.redirect(`${url.protocol}//${url.host}/blocked`)
@@ -35,7 +37,7 @@ export async function middleware(request: NextRequest) {
     const isFromVpn = await isVPN(ip)
     if (isFromVpn && url.pathname !== '/blocked') {
       try {
-        await redis.set(ip, 1)
+        await redis.set(ip, BLOCKED_IP_VALUE)
       } catch (error) {
         console.error('Failed to set data in Redis:', error)
       }
--- a/packages/frontend/middleware.ts
+++ b/packages/frontend/middleware.ts
@@ -9,6 +9,8 @@ const redis = new Redis({
   token: process.env.UPSTASH_REDIS_REST_TOKEN!,
 })
 
+const THIRTY_DAYS_IN_MS = 30 * 24 * 60 * 60 * 1000
+
 interface RedisResponse {
   value: string
   timestamp: number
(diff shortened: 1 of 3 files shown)


packages/frontend/middleware.ts
```diff
@@ -19,17 +21,16 @@ export async function middleware(request: NextRequest) {
 
   if (ip && !isIPWhitelisted) {
     const redisData = await redis.get(ip)
-    const isIPBlocked = !!redisData
+    const isIPBlocked = redisData === BLOCKED_IP_VALUE
```
nikkaroraa, line 24: older values don't make any sense now, similar to refreshing the older blocked ips list

aleone: Yeah thats fair.  

The only reason for the storage of these was to reduce risk of hitting api limits on ipqs.  

In future may want to have a reset of values at some cadence and store the timestamp when it is saved to handle changes in ip ranges/data centers/vpns etc. Something like this below terrible pseudocode:

``` isIPBlocked redisData[0] === BLOCKED_IP_VALUE && (Date.now() - redisData[1]) / 86400000< 30 ```

then later just setting 2 values, maybe array or something, not super familiar on what possible there, to redis:

``` await redis.set(ip, [BLOCKED_IP_VALUE, Date.now()]) ```

File: packages/frontend/src/constants/index.ts [hunk 2]
```diff
@@ -70,6 +70,9 @@ export const BLOCKED_COUNTRIES = [
 ]
 export const ALLOW_WITHDRAWALS = ['CA']
 
+export const WHITELISTED_DATACENTER_ORGANIZATIONS = ['amazon', 'google', 'azure', 'oracle', 'softlayer', 'cloudflare']
```
aleone, line 73: should this be ENV for easier adjustment, etc?

aleone (approved):
generally lgtm

1 comment on future re resetting redis db + storing a time stamp

alpinechicken (commented):
lgtm - might be good to do a test to check if gcp or aws can get through

packages/frontend/middleware.ts
nikkaroraa: yeah agreed thats better. i'll make another PR to implement this.

packages/frontend/src/constants/index.ts
nikkaroraa: agreed

packages/frontend/src/server/ipqs.ts
```diff
@@ -62,17 +64,27 @@ async function get_IPQ_URL(url: string): Promise<ApiResponse | null> {
...
-    return true
-  } else {
-    return false
+  const isSuspectedToBeVPN = ipResult.vpn === true || ipResult.tor === true
+  if (isSuspectedToBeVPN) {
+    const org = ipResult.organization
+
+    // exclude datacenter IPs
+    // organization should include one of WHITELISTED_DATACENTER_ORGANIZATIONS
+    const isWhitelistedOrg = WHITELISTED_DATACENTER_ORGANIZATIONS.some((whitelistedOrg) =>
```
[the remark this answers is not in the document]
nikkaroraa: you can hardcode the ipAddress to be one from [this sheet](https://docs.google.com/spreadsheets/u/1/d/1vzXb1u4cQlpzKCcyyl44HAac8ay6to2wpDShgZVF-hQ/edit#gid=0)

packages/frontend/src/constants/index.ts
nikkaroraa: done: https://github.com/opynfinance/squeeth-monorepo/pull/1012/commits/e075f27bd58e58de5711df971f55114375abfa1a

packages/frontend/middleware.ts
nikkaroraa: implemented it here: https://github.com/opynfinance/squeeth-monorepo/pull/1015

alpinechicken (approved):
changes lgtm

PR #1015, opened 2023-12-05
feat: expiry based vpn cache

nikkaroraa:
# Task:
Added timestamp to check if the VPN cache entry isn't too old. Currently the deadline is set to 15 days, so all the entries before 15 days won't be considered.

## Description

Fixes ENG-1970

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Testing code


alpinechicken (approved):
lgtm

PR #1018, opened 2024-01-29
reduce cache expiry time to 3 days

nikkaroraa:
# Task:

Reduce cache expiry time from 30 days to 3 days

## Description

Originally, the setting was for 30 days, indicating that when a data center is whitelisted, existing users would only gain access to the site after 30 days from their initial visit. However, a 3-day period seems to be an optimal balance, reducing the frequency of API calls while also avoiding issues with outdated data.

Fixes ENG-2043

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Testing code

### DIFF
--- a/packages/frontend/middleware.ts
+++ b/packages/frontend/middleware.ts
@@ -9,7 +9,7 @@ const redis = new Redis({
   token: process.env.UPSTASH_REDIS_REST_TOKEN!,
 })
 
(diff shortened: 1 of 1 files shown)


alpinechicken (approved):
lgtm - why is tests and vercel failing?

nikkaroraa:
> lgtm - why is tests and vercel failing?

zenbull stuff. gotta remove those

PR #1034, opened 2024-08-06
FEAT: Terms of Service & Privacy Policy Change <ENG-2538>

0xMuTwo:
# Task:

`/terms-of-service` and `/privacy-policy` routes are hidden behind the location blocker. 
## Description

Alterations to restricted blocker for `/terms-of-service` and `/privacy-policy`.

Fixes (ENG-2538)

## Type of change

- [ ] New feature
- [X] Bug fix
- [ ] Testing code

### DIFF
--- a/packages/frontend/src/components/StrikeWarning.tsx
+++ b/packages/frontend/src/components/StrikeWarning.tsx
@@ -105,7 +105,7 @@ export default function StrikeModalManager() {
     }
   }, [isRestricted, isExcludedPage])
 
-  if (showModal && !isExcludedPage) {
+  if (showModal) {
     if (isWalletConnected) {
       return <StrikeCountModal setIsStrikeModalShownOnce={setIsStrikeModalShownOnce} />
     }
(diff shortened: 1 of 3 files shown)


nikkaroraa (approved):
looks good. lets merge it 🚀

packages/frontend/src/components/StrikeWarning.tsx
```diff
@@ -89,18 +90,22 @@ export default function StrikeModalManager() {
...
+  const router = useRouter()
+
+  // Check if the current page is either terms of service or privacy policy
+  const isExcludedPage = ['/terms-of-service', '/privacy-policy'].includes(router.pathname)
 
   // delayed state change - show modal after 2 seconds
   // this is because it takes some time to automatically connect the wallet (incase previously connected)
   useEffect(() => {
-    if (isRestricted) {
+    if (isRestricted && !isExcludedPage) {
```
nikkaroraa, line 101: i don't think we need to check `isExcludedPage` here. 

`if (showModal && !isExcludedPage)` this should be enough

nikkaroraa (changes requested):

packages/frontend/src/components/StrikeWarning.tsx
0xMuTwo: Awesome catch, you're right this is redundant. Fixed in new version.

nikkaroraa (approved):
looks good - merging

